PR #2, opened 2018-07-11
curtin: Test upstream/master with Ubuntu proposed enabled.

smoser:
Instead of N jobs (one per ubuntu release) running vmtest to test
ubuntu -proposed, just run a single vmtest using master.
The alternative has some value, but I think we get much more bang for
the buck this way.

a.) with the skip_by_date decorator, we'll see if using proposed
actually fixes any issues. If we do SKIP_BY_DATE_BUGS=* we simply skip
those bugs (and that is needed there).

b.) we only have to have one job, not 4.
we get the same coverage with the one job, as it runs curtin installs
of each supported release.  The 4 jobs don't actually change the
coverage at all as the thing intended to be tested here is Ubuntu
-proposed, rather than curtin itself.

c.) we can fix any issue found quickly with fixes to trunk and not have
to wait for SRU.

Also here, enable setting repos in the curtin-vmtest-devel-amd64
test.


curtin/jobs-vmtest.yaml
```diff
@@ -108,6 +109,24 @@
...
+        - vmtest-add-repos
+    properties:
+      - build-discarder:
+          num-to-keep: 5
+    publishers:
+        - email-server-crew
+        - archive-results
+    builders:
+      - curtin-vmtest:
+          vmtest-add-repos: proposed
```
smoser: this syntax is wierd to me. i'm not sure if it is correct, the intent is to provide a value for that variable here, and that this value override the default (empty) value that is used everywhere else.

curtin/jobs-vmtest.yaml
```diff
@@ -98,6 +98,7 @@
         - landing-candidate
         - landing-candidate-branch
         - nose-args
+        - vmtest-add-repos
```
smoser: the intent here is to let the user starting a devel-debug build provide this variable if they are so inclined but if not it would get the default "" value.

powersj:
Because you are passing vmtest-add-repos as a parameter there is no need to add it as a builder sub-key

This will get the tests to pass:
https://paste.ubuntu.com/p/KCb9CP3yxP/

smoser:
@powersj 

> Because you are passing vmtest-add-repos as a parameter there is no need to add it as a builder sub-key

I'm not sure... the default value of that parameter is "" (or, otherwise unset).  I just wanted to set the value for this job.

powersj:
Closing as we have a solution.

PR #3, opened 2018-07-17
simplestreams: Move to using git and run tox in a container.

smoser:
Simplestreams has moved from bzr to git, and bzr will stop working
shortly. [1]

It also changes the tests to run in an lxc (bionic) container via
utilization of 'ctool' which was added today to uss-tableflip.

Hopefully running in a container will fix the errors we've been
seeing lately around gpg:
  gpg: can't connect to the agent: IPC connect call failed
  gpg: error getting the KEK: No agent running

[1] https://code.launchpad.net/~smoser/simplestreams/trunk.moved-to-git/+merge/348547

### DIFF
--- a/simplestreams/jobs.yaml
+++ b/simplestreams/jobs.yaml
@@ -54,12 +54,16 @@
           #!/bin/bash -x
           export https_proxy=http://squid.internal:3128
           export no_proxy=launchpad.net
-
           rm -rf *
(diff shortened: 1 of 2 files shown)


powersj:
ctool is pretty cool :)

fwiw simplestreams is already using git. We use a git config that translates lp:simplestreams to the correct URL.

PR #6, opened 2018-07-25
simplestreams: use ctool for c-i review.

smoser:
This just makes the simplestreams c-i review tool use ctool as
we had done with the trunk tests.  We need to do that for the same reason
as the previous... to avoid gpg agent issues
 gpg: can't connect to the agent: IPC connect call failed

### DIFF
--- a/simplestreams/jobs-ci.yaml
+++ b/simplestreams/jobs-ci.yaml
@@ -73,12 +73,13 @@
           set -eux
           export https_proxy=http://squid.internal:3128
           export no_proxy=launchpad.net
-
           rm -rf *
-          git clone ${landing_candidate} ss-${BUILD_NUMBER}
-          cd ss-${BUILD_NUMBER}
-          git rev-parse HEAD
-          git remote add upstream https://git.launchpad.net/simplestreams
-          git fetch upstream --tags
 
-          tox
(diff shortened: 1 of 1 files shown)


simplestreams/jobs-ci.yaml
```diff
@@ -73,12 +73,13 @@
...
-          git remote add upstream https://git.launchpad.net/simplestreams
-          git fetch upstream --tags
 
-          tox
+          git clone https://github.com/CanonicalLtd/uss-tableflip.git uss-tableflip
+          PATH=$PWD/uss-tableflip/scripts:$PATH
+
+          ctool run-container -v --destroy "--name=${JOB_NAME}-$BUILD_NUMBER" \
+              ubuntu-daily:bionic \
+              --git="{landing_candidate}@{candidate_revision}" \
```
powersj: landing_candidate will come in the format of lp:~powersj/simplestreams/test, is the container's git setup to handle those type of strings? otherwise we should probably reformat it.

smoser: i had noticed that, and just added it to ctool:
 https://github.com/CanonicalLtd/uss-tableflip/commit/e3743ea31021b492122c35315af419191be2ca69

PR #9, opened 2018-08-08
curtin: temporarily enable keep_data_fail and tar

powersj:
These options will enable the taring up of disks and keeping them around when a specific test fails. This was requested by @raharper for a temporary test.


raharper:
Thanks!  Looks correct.

PR #10, opened 2018-08-08
curtin: temporarily enable keep_data_fail

powersj:
Enable keeping of disks when a specific test fails. This was requested by @raharper for a temporary test.


raharper:
Thanks!  Looks correct.

PR #11, opened 2018-09-20
switch foundations_sru from prometheus to influxdb

murraybd:
switch the job since we switch the metric

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -41,18 +41,6 @@
           python3 -m metrics.{metric_module} {metric_arguments}
 
 - builder:
-    name: errors-most-common-problems
-    builders:
-      - shell: |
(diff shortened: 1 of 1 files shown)


metrics/metrics.yaml
```diff
@@ -130,7 +130,8 @@
 - job:
     name: metric-foundations-sru-queues
     builders:
-        - simple-metric:
+        - metric-influxdb:
+            team: "foundations"
             metric_module: "foundations_sru"
```
powersj: You also need a "metric_arguments" parameter like:

`metric_arguments: ""`

This is because the runner allows us to pass parameters to the script and jenkins job builder needs all the parameters filled in.

murraybd: Ah, I didn't realize it was required.

PR #12, opened 2018-09-24
Use the metric-influxdb builder for metric-foundations-proposed-migration.

sil2100:


sil2100:
Required: https://github.com/CanonicalLtd/metrics/pull/8

PR #14, opened 2018-09-25
chmod non-writable dirs in vmtest output dir

raharper:
Centos vmtests sometimes leave a root dir with non-writable
permissions which prevents the rm -Rf from cleaning up.


powersj:
This has conflicts and other issues, so I'm going to close it. If we need to revisit, feel free to open again

PR #16, opened 2018-10-11
cloud-init: Match test harness version and daily deb version.

smoser:
When running integration tests, we should match the version of the
test-harness with the version of the deb we're running.  Otherwise
we will get skew and have test failures when a test makes an assertion
that is not availble in the installed package or vice-versa.

The fix here pulls the hash out of the daily deb file name.
   cloud-init_18.4-1964-g9f88125-0ubuntu1+1635~trunk~ubuntu18.04.1_all.deb

It grabs the -g9f88125 there and then checks out cloud-init at gf88125.

### DIFF
--- a/cloud-init/integration-ec2.yaml
+++ b/cloud-init/integration-ec2.yaml
@@ -64,17 +64,28 @@
     name: integration-ec2
     builders:
       - shell: |
+          set -e
           release="{release}"
 
           sudo rm -Rf *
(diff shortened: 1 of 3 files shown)


smoser:
This is the simplist answer I could come up with.  it should work for the daily ppa, and any cloud-init that has a 'gXXXXXX' in its filename.   That is expected to be any daily-deb version at this point.

PR #17, opened 2018-10-11
curtin: Add a vmtest_filters parameter for proposed testing default to ubuntu only

raharper:
We want to only test ubuntu for our proposed testing.  Add a new
parameter to the proposed test and set the default value to
target_distro=ubuntu.  This allows users to customize the job
later so add additional (or remove) the filter as well.

### DIFF
--- a/curtin/jobs-vmtest.yaml
+++ b/curtin/jobs-vmtest.yaml
@@ -149,10 +149,10 @@
             if [ -n "${vmtest_filters}" ]; then
                 f=""
                 for f in ${vmtest_filters}; do
-                    filter_args="${filter_args} --filter ${f}"
+                    filter_args="$filter_args --filter=$f"
                 done
             fi
-            echo "test filter args: \"${filter_args}\""
+            echo "test filter args: \"$filter_args\""
 
             # Helpful in debugging, but fill up Jenkins if all test fail
             # export CURTIN_VMTEST_KEEP_DATA_FAIL=all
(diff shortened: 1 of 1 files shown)


curtin/jobs-vmtest.yaml
```diff
@@ -121,6 +121,9 @@
         - string:
             name: vmtest_add_repos
             default: "proposed"
+        - string:
+            name: vmtest_filters
+            default: "target_distro=ubuntu"
```
smoser: we could just set the default nose_args for this test to include `--filter=target_distro=ubuntu` right? rather than adding a new variable ?

curtin/jobs-vmtest.yaml
```diff
@@ -140,11 +143,21 @@
...
                 export CURTIN_VMTEST_ADD_REPOS=${vmtest_add_repos}
             fi
-            
+
+            # construct vmtest filters
+            filter_args=""
+            if [ -n "${vmtest_filters}" ]; then
+                f=""
+                for f in ${vmtest_filters}; do
+                    filter_args="${filter_args} --filter ${f}"
```
smoser: filter_args="${filter_args} --filter=$f"
Ie, use --filter=X rather than --filter X
its much more clear.

smoser:
this looks fine to me.

smoser:
Just as later reference, there is a bunch of conversation in irc logged at https://irclogs.ubuntu.com/2018/10/12/%23snappy.html#t13:49 .

PR #20, opened 2018-11-13
use --ignore-cache on jenkins-jobs update

blackboxsw:


blackboxsw:
> A few questions:
> 
> 1. Why not always use '--ignore-cache' ?  Is that a performance thing?  Do we care? If this is a difference between something taking 1 second and 12 seconds and 12 seconds is fast enough... lets just go with always using it.
Answer: Runtime is ~2 mins to force update all, 21 seconds to only update based on observed cache delta

> 2. If we decide we want to use the cache and not always pass --ignore-cache, then would it be enough to just make it a boolean?  If you make the type a boolean, then there is a check box and the user doesn't have to knwo what value they can put there.  As it is right now you have to know to use 'ALL' and not 'all' or '*' ...

If we make it a boolean, then this will solve most normal use-cases:
  -  where we expect to update the jobs automaticall only when we have pushed a new commit to server-jenkins-jobs repo
  - set --force-update parameter True if we have accidentally or intentionally altered a job's configuration in the UI and want all jobs to be fixed to match tip of server-jenkins-jobs

PR #25, opened 2019-01-28
cloud-init: set http_proxy and https_proxy before calling run-centos

paride:
Another step towards fixing `cloud-init-copr-build`.

### DIFF
--- a/cloud-init/centos.yaml
+++ b/cloud-init/centos.yaml
@@ -35,7 +35,11 @@
           git clone https://github.com/CanonicalLtd/server-test-scripts
           git clone https://git.launchpad.net/cloud-init
           cd cloud-init
+
           export PYVER=python2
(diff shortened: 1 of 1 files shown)


powersj:
Tested this on Torkoal?

cloud-init/centos.yaml
```diff
@@ -34,8 +34,12 @@
           
           git clone https://github.com/CanonicalLtd/server-test-scripts
           git clone https://git.launchpad.net/cloud-init
-          cd cloud-init
```
raharper: No need to move the 'cd cloud-init';

paride:
@powersj yes, works fine on torkoal, however completely fixing cloud-init-copr-build will require a couple of changes to https://github.com/paride/server-test-scripts/ too.

@raharper amended, and now the commit is done using my @canonical.com address.

PR #28, opened 2019-01-30
Remove the metric-server-google-analytics job

paride:
Setup long ago at an attempt to collect metrics from it, but not doing
anything useful.


powersj:
thanks!

PR #29, opened 2019-01-31
Improve the failure email sending logic

paride:
Always use the email-server-crew publisher when appropriate.

### DIFF
--- a/cloud-init/centos.yaml
+++ b/cloud-init/centos.yaml
@@ -21,8 +21,7 @@
     triggers:
       - timed: "H 23 * * *"
     publishers:
-      - email:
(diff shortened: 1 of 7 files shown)


paride:
Tested on torkoal with `jenkins-jobs test`.

paride:
I didn't add the email publisher to the CI jobs running every 15 minutes, e.g. `cloud-init-ci-trigger`, `cloud-init-autoland-trigger` and the equivalents for curtin. We have 6 of such jobs, and problems are normally easy to spot as they break the CI (and thus our development workflow), so receiving a deluge of error emails is probably not necessary, but let me know what you think.

PR #34, opened 2019-02-05
add in a job for the new foundations_proposed_migration_by_team metric

murraybd:


powersj:
Can you add the job name to default.yaml as well please?

murraybd:
> Can you add the job name to default.yaml as well please?

I felt like I'd forgotten something. Thanks!

PR #35, opened 2019-02-21
Add metric for distro MIR bug counts

cyphermox:


metrics/metrics.yaml
```diff
@@ -275,3 +275,10 @@
           team: "server"
           metric_module: "uploads"
           metric_arguments: "openstack"
+
+- job:
+    name: metric-distro-mir-bugs
+    builders:
+      - metric-influxdb:
+          team: "foundations"
+          metric_module: "mir_team_bugs"
```
murraybd: I think metric_arguments is required but it can just be empty.

PR #39, opened 2019-03-18
Add in jobs to create metrics of retracing results

murraybd:
For both staging and production environments

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -283,3 +283,19 @@
           team: "foundations"
           metric_module: "mir_team_bugs"
           metric_arguments: ""
+
+- job:
+    name: metric-foundations-retracers-results-production
+    builders:
+      - metric-influxdb:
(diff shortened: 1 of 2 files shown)


metrics/metrics.yaml
```diff
@@ -283,3 +283,19 @@
           team: "foundations"
           metric_module: "mir_team_bugs"
           metric_arguments: ""
+
+- job:
+    name: metric-foundations-retracers-results-production
```
OddBloke: If I recall correctly from the metric script, these only need to run once per day. Is it worth adding:

```yaml
triggers:
  - timed: "0 H * * *"
```

so that we only run these jobs once per day?

murraybd: I didn't know that "timed" was a possible parameter. That being said, while I only want to store the final number for a day I'd like to be able to check how successful retraces are during the day.

PR #42, opened 2019-03-25
add timeout to cloud-init,curtin,simplestreams trigger jobs

OddBloke:
These run in a predictable amount of time, so if they're taking more
than 15 minutes then we've almost certainly hit some sort of issue.

### DIFF
--- a/cloud-init/jobs-ci.yaml
+++ b/cloud-init/jobs-ci.yaml
@@ -20,6 +20,10 @@
     node: metal-amd64
     triggers:
         - timed: H/15 * * * *
+    wrappers:
+        # We run every 15 minutes, and runs usually take ~7 minutes
+        - timeout:
+            timeout: 15
     builders:
         - shell: |
             #!/bin/bash -x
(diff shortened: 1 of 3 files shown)


OddBloke:
This may also require installation of the timeout plugin on our Jenkins instance; Jenkins will ignore this configuration if there isn't a plugin that uses it, so we can safely merge this whilst we resolve that.

paride:
@powersj Is the plugin installation to be requested with a RT?

powersj:
@paride yes, one of you will need to submit the RT, please CC me

paride:
I will, as I'd like to request the "timestamp" plugin" too.

OddBloke:
On Mon, Mar 25, 2019 at 08:11:18AM -0700, Paride Legovini wrote:
> I will, as I'd like to request the "timestamp" plugin" too.

Thanks!

paride:
Anyway LGTM, I'll go ahead and merge it.

PR #43, opened 2019-04-22
curtin: reschedule vmtests add disco

raharper:
Curtin vmtest load is high and current runs take about 14 hours.
This branch does:

- Add disco to daily and -proposed
- Daily drops trusty, adds bionic, reschedules to 4 UTC, xenial
  on days 1,3,5, and bionic on days 2, 4, 6.
- Proposed splits up xenial -> disco over different days.
- Start vmtest-devel-amd64-proposed a bit later.

### DIFF
--- a/curtin/default.yaml
+++ b/curtin/default.yaml
@@ -43,7 +43,6 @@
       - curtin-vmtest-proposed-d
       - curtin-vmtest-proposed-d-trigger
       - curtin-vmtest-proposed-x
-      - curtin-vmtest-proposed-x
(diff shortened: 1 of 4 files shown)


raharper:
Please review the timed changed, I think I'm doing that right.

powersj:
That looks good to me, but you need to update the master list of jobs here:
https://github.com/CanonicalLtd/server-jenkins-jobs/blob/master/curtin/default.yaml#L18

Then CI will pass

raharper:
Thanks, pushed a fix

PR #45, opened 2019-05-07
New job: daily cloud-init/curtin bug triage email

paride:


paride:
Please have a look at the email this jobs sends while reviewing :)

paride:
The project is called `daily-triage` (without reference to curtin or cloud-init) as it is now easy expand the job to send an email for the Squeaky Wheel triage too.

daily-triage/daily-triage.yaml
```diff
@@ -0,0 +1,157 @@
...
+- job:
+    name: cloud-init-curtin-daily-triage
+    node: torkoal
+    triggers:
+      - timed: "H(0-20) 6 * * 1-5"
+    publishers:
+      - email-server-crew
+    builders:
+      - shell: |
+          #!/bin/bash -ex
```
OddBloke: Given its size, I wonder if it's worth having this shell script in a separate file and using `!include-raw` ([docs](https://docs.openstack.org/infra/jenkins-job-builder/definition.html?highlight=include#inclusion-tags)) to include it in our job definition?

paride: Thanks @OddBloke, yes it makes sense, let me update the PR. The script grew quite a lot.

PR #50, opened 2019-05-21
cloud-init: separate proposed platform jobs

blackboxsw:
Jenkins *proposed jobs are long-running and frequently get hit by intermittent errors in either lxd (teardown) or kvm (gpg key/timeout errors). Decoupling these tests from one another based on platform will hopefully give us a chance to see green in one platform at at time and avoid having to re-run a two hour test.

### DIFF
--- a/cloud-init/default.yaml
+++ b/cloud-init/default.yaml
@@ -42,11 +42,21 @@
       - cloud-init-integration-nocloud-kvm-d
       - cloud-init-integration-nocloud-kvm-e
       - cloud-init-integration-nocloud-kvm-x
-      - cloud-init-integration-proposed-b
-      - cloud-init-integration-proposed-c
-      - cloud-init-integration-proposed-d
-      - cloud-init-integration-proposed-e
-      - cloud-init-integration-proposed-x
+      - cloud-init-integration-proposed-b-ec2
+      - cloud-init-integration-proposed-b-lxd
+      - cloud-init-integration-proposed-b-kvm
+      - cloud-init-integration-proposed-c-ec2
+      - cloud-init-integration-proposed-c-lxd
+      - cloud-init-integration-proposed-c-kvm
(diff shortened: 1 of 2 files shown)


blackboxsw:
Kicked off individual test runs on jenkins to validate subcomponents:
 https://jenkins.ubuntu.com/server/view/cloud-init/job/cloud-init-integration-proposed-d-lxd/1
 https://jenkins.ubuntu.com/server/view/cloud-init/job/cloud-init-integration-proposed-d-ec2/1
 https://jenkins.ubuntu.com/server/view/cloud-init/job/cloud-init-integration-proposed-d-nocloud-kvm/1

OddBloke:
IMO it's useful to have a single job that indicates at a high-level whether things are good or not; I _think_ we lose that here (though my internal JJB parser may be failing me). Could we use a matrix job with a child to trigger each of the separate platform jobs, so that the parent job would capture that high-level status?

Another thought: assuming these tests run serially now, will running all these tests in parallel make them more prone to failure? Probably not, because they're all different platforms, but I thought I'd ask because I don't know enough to make my own judgement.

powersj:
Reasons for splitting this up include not having to wait 2 hours in the event we want to re-run tests, the ability to re-run a single platform, have faster execution to do smoke testing of proposed (e.g. run lxd tests).

We have at least two external issues causing failures currently: 1) key servers not being reliable 2) lxd zfs delete issues. By breaking this up, it makes it easier to see legit failures versus non-legit and the ability to more easily re-run tests.

I do agree that you sort of loose the high-level status, but since these are for proposed, which requires us manually submitting the results I am less concerned about that since we will have to manually review anyway.

This also moves the proposed tests back to being more in line with the daily tip tests, which run by platform.

OddBloke:
Yep, all good from my POV. Thanks all!

PR #52, opened 2019-06-05
docker: test the latest Ubuntu docker image size

paride:


docker/default.yaml
```diff
@@ -0,0 +1,56 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+
+- defaults:
+    name: global
+    description: |
+        Curtin Ubuntu Server QA Jenkins Job
```
OddBloke: This needs changing, I think?

docker/docker-latest-image-size.sh
```diff
@@ -0,0 +1,29 @@
...
+lxc launch ubuntu: docker-latest-image-size-test -e
+sleep 5
+lxc exec docker-latest-image-size-test -- apt update
+lxc exec docker-latest-image-size-test -- apt -y install docker.io
+lxc exec docker-latest-image-size-test -- mkdir /etc/systemd/system/docker.service.d
+lxc exec docker-latest-image-size-test -- \
+	sh -c 'echo "[Service]\nEnvironment=\"HTTPS_PROXY=http://squid.internal:3128/\"" > /etc/systemd/system/docker.service.d/https-proxy.conf'
+lxc exec docker-latest-image-size-test -- systemctl daemon-reload
+lxc exec docker-latest-image-size-test -- systemctl restart docker
+lxc exec docker-latest-image-size-test -- docker pull ubuntu:latest
```
OddBloke: Should this be `ubuntu:devel` so we're testing eoan images? Or are we intentionally testing the latest LTS?

(See https://hub.docker.com/_/ubuntu for details of which tag means what.)

paride: Let's have three jobs for `latest`, `rolling` and `devel`.

paride:
I'd like to have the size limits are Jenkins parameters too, so I'm going to update this PR once again.

paride:
Ready for review!

docker/image-size.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+- job-template:
+    name: docker-{dockertag}-image-size
+    node: metal-amd64
+    parameters:
+      - string:
+          name: DOCKERTAG
+          default: '{dockertag}'
+          description: "Ubuntu docker tag to pull."
+      - docker-size-limit-{dockertag}
```
OddBloke: I _think_ you could remove the need for the separate parameter definitions by doing something along the lines of:

```yaml
- string:
    name: MAXSIZE
    default: !j2: '{{ {'latest': 72, 'rolling': 80, 'devel': 80}[dockertag] }}
```

(`!j2` treats its argument as a Jinja2 template.)

paride: Thanks, I *knew* there was a way! Unfortunately it is not working:

```
yaml.constructor.ConstructorError: could not determine a constructor for the tag '!j2:'
```

but the JJB on torkoal it at version 2.0.2, it could be too old. I'll try with a newer version.

OddBloke: Haha, only slightly too old, it was introduced in 2.0.3.

paride:
It works!

OddBloke:
image

docker/docker-image-size.sh
```diff
@@ -0,0 +1,40 @@
...
+if lxc list --format csv -c n | grep -q "^${container}$"; then
+    lxc delete "$container" --force
+    sleep 5
+fi
+
+lxc launch ubuntu: "$container" -e
+
+# Networking is not *immediately* ready in the newly created container.
+# Let's sleep for a few seconds.
+sleep 10
```
OddBloke: Could this be replaced with something along the lines of `lxc exec "$container" -- cloud-init status --wait`?

paride: Definitely a better idea; done. I would have preferred to use `systemctl is-system-running --wait`, which should be even safer, but:
* `--wait` is not implemented in Bionic's systemd
* Calling `systemctl is-system-running` too early causes a `Failed to connect to bus: No such file or directory` error, which kind of defeats the purpose of the subcommand. This is reproducible like this:

```
$ lxc launch ubuntu-daily:e paride-delme ; lxc exec paride-delme -- systemctl is-system-running --wait                                                                            
Creating paride-delme
Starting paride-delme
Failed to connect to bus: No such file or directory
```

e.g. on diglett.

PR #60, opened 2019-07-03
explicitly set 'fail: true' in the timeout wrappers

paride:
The JJB default is to take no action if the timeout is reached; the
action has to be set explicitly.


OddBloke:
TIL!

PR #61, opened 2019-07-09
daily-triage email: add simplestreams

paride:
Changes:
 - Cleanup the workspace before each run
 - On failures email the QA engineer only
 - Add simplestreams

### DIFF
--- a/daily-triage/default.yaml
+++ b/daily-triage/default.yaml
@@ -34,8 +34,19 @@
       - build-discarder:
           num-to-keep: 30
 
+- builder:
+    name: cleanup-workspace
+    builders:
+      - shell: "find $WORKSPACE -mindepth 1 -delete"
+
 - publisher:
     name: email-server-crew
     publishers:
       - email:
           recipients: server-crew-qa@lists.canonical.com
+
+- publisher:
+    name: email-qa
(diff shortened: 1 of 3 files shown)


daily-triage/default.yaml
```diff
@@ -34,8 +34,19 @@
       - build-discarder:
           num-to-keep: 30
 
+- builder:
+    name: cleanup-workspace
```
OddBloke: Why not https://docs.openstack.org/infra/jenkins-job-builder/wrappers.html#wrappers.workspace-cleanup%20(pre-build) ?

paride: The only reason is that the plugin is not installed, or at least I don't think it is, as I didn't ask for it. I'll file a RT requesting it in all our Jenkins.

paride: And I find slightly annoying that if the plugin is not installed nothing will happen: `jenkins-jobs update` will succeed and that's it.

OddBloke: That's a reasonable response; I'm happy keeping it as-is.

paride:
I asked for the Workspace Cleanup plugin to be installed in out Jenkins.

PR #67, opened 2019-07-31
docker-image-size: improve script to avoid leaving stale containers

paride:
Changes:
 - trap EXIT and call a container cleanup function
 - use printf(1) instead of echo to print strings with backslash escapes
 - use apt-get instead of apt, following the advice given by apt(8)
 - cosmetic changes when running with `set -x`

While this version of the script is unlikely to leave stale containers,
as the `cleanup` function is normally called twice (the first time
explicitly, the second time by the trap), it will still exit nonzero on
"ZFS dataset is busy" issues. We still don't want to work around them,
as a proper fix is needed.

### DIFF
--- a/docker/docker-image-size.sh
+++ b/docker/docker-image-size.sh
@@ -13,30 +13,31 @@ set -eufx -o pipefail
 
 container="docker-$DOCKERTAG-image-size-test"
 
-# If there is a stale container delete it
-if lxc list --format csv -c n | grep -q "^${container}$"; then
-    lxc delete "$container" --force
-    sleep 5
-fi
+function cleanup {
+    # Is there a better way to check if a given LXD container exists?
+    ! lxc info "$container" >/dev/null 2>&1 || lxc delete "$container" --force
+}
 
+trap cleanup EXIT
+
+cleanup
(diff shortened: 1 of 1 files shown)


docker/docker-image-size.sh
```diff
@@ -13,30 +13,31 @@ set -eufx -o pipefail
...
 lxc exec "$container" -- systemctl restart docker
 lxc exec "$container" -- docker pull "ubuntu:$DOCKERTAG"
 size=$(lxc exec "$container" -- docker image ls "ubuntu:$DOCKERTAG" --format "{{.Size}}")
-lxc delete "$container" --force
+cleanup
 
 size=$(echo "$size" | grep -o '^[0-9]*')
-echo "Maximum allowed size: ${MAXSIZE}MB"
-echo "Current size: ${size}MB"
+: "Maximum allowed size: ${MAXSIZE}MB"
```
OddBloke: What effect does this replacement have? Avoids logging the message twice because of `-x`?

paride: Yes exactly. Makes the output more readable IMHO.

PR #70, opened 2019-08-05
git-ubuntu-ci: use double quotes when setting VM_NAME

paride:
Apparently in newer Jenkins versions variables set with single quotes
are not expanded anymore. Switch to double quotes.


panlinux:
+1

PR #72, opened 2019-08-21
Run the CI jobs on any metal-amd64 node

paride:
Run the CI jobs on any metal-amd64 node instead of forcing the jobs to
run on torkoal.

### DIFF
--- a/cloud-init/jobs-ci.yaml
+++ b/cloud-init/jobs-ci.yaml
@@ -67,7 +67,7 @@
     sandbox: True
     dsl: |
       pipeline {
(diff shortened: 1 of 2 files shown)


OddBloke:
Do all the `metal-amd64` hosts have the dependencies we assume to be present installed?

paride:
Ensuring we have the toolchain installed and in sync on all the nodes is what I've been doing the last half an hour :) So hopefully yes.

OddBloke:
Let's see how well you did! ;-)

PR #77, opened 2019-09-09
Delete the curtin-vmtest-sync-images job

paride:
The vmtests already sync the needed images before running, so this job
is not really needed.

As we plan to run the vmtests across a number of arches a single job
wouldn't have worked well, as Jenkins runs it on a single node, while
we'd like to update the images on all the nodes.

### DIFF
--- a/curtin/default.yaml
+++ b/curtin/default.yaml
@@ -33,7 +33,6 @@
       - curtin-vmtest-devel-amd64-proposed
       - curtin-vmtest-devel-ppc64el
       - curtin-vmtest-devel-s390x
-      - curtin-vmtest-sync-images
       - curtin-vmtest-daily-x
       - curtin-vmtest-daily-b
       - curtin-vmtest-proposed-b
--- a/curtin/jobs-vmtest.yaml
+++ b/curtin/jobs-vmtest.yaml
@@ -16,20 +16,6 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
 - job:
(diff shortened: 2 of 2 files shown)


paride:
It seems to me that the vmtests are perfectly able to download the images they need. Do we really need this job?

If we do, then we need to change how it works, as a single job is not suitable to update the images across a number of nodes. We could for example convert it to a JJB builder, and run it before each vmtest run. This would have two disadvantages:

 - Only one vmtest job would run on each node at a given time, as the sync-images script has a locking mechanism that checks if `jenkins-runner` is running
 - The images would get synced very often (but no actual data is downloaded if they're already in sync).

The sync-images builder could check the timestamp of the most recent file in `/srv/images` and `exit 0` if that's less than e.g. 12 hours ago. But again: do we need to run it at all?

powersj:
There is history here that you are missing :)

We typically have multiple vmtests running at the same time. If these tests run and update images, while another test is running then the older run will start failing as it will not be able to find the images it was originally using.

We update images using a specific job to prevent the above from happening and to sync the images once per day by only a single job. Yes sometimes it gets delayed, but it prevents tests from stomping on each other.

If you want vmtest to sync the images on arm64 and ppc64 then that is probably ok to start with, but if we add any other tests we will run into this.

paride:
I was expecting something like this :) And now I see that `tools/jenkins-runner` disables the automatic image sync done by the vmtests.

No need to say it but: do not approve/merge this for the moment, I want to think about it a bit more and come up with a good solution.

PR #79, opened 2019-09-13
Enable curtin-vmtest-devel-ppc64el

paride:
Run only test_basic.py for the moment.


paride:
It took me a while to realize that if *no* parameter is passed to a macro then no substitution is done. This is why we need to always pass `default_nose_args`, even if empty.

https://docs.openstack.org/infra/jenkins-job-builder/definition.html#macro-notes

PR #81, opened 2019-10-01
Replace the curtin-vmtest-sync-images job with a builder

paride:
The builder is run before each vmtest run, checks if the images have been
updated recently enough and runs vmtest-sync-images if necessary. This
will update the images on any node the vmtests are run on.

The builder is not run as part of the curtin-ci job, as we want it to
run as soon as possible, with the existing images.

### DIFF
--- a/curtin/jobs-vmtest-daily.yaml
+++ b/curtin/jobs-vmtest-daily.yaml
@@ -30,7 +30,6 @@
         - email-server-crew
         - archive-results
     builders:
-      - curtin-vmtest-sync-images
       - vmtest-daily:
           release: xenial
           nose_args: nose_args
@@ -50,7 +49,6 @@
         - email-server-crew
         - archive-results
     builders:
-      - curtin-vmtest-sync-images
       - vmtest-daily:
           release: bionic
           nose_args: nose_args
@@ -59,6 +57,7 @@
 - builder:
     name: vmtest-daily
     builders:
+      - curtin-vmtest-sync-images
       - shell: |
           RELEASE={release}
           TYPE="daily"
--- a/curtin/jobs-vmtest-daily.yaml
+++ b/curtin/jobs-vmtest-daily.yaml
@@ -30,6 +30,7 @@
         - email-server-crew
         - archive-results
     builders:
+      - curtin-vmtest-sync-images
       - vmtest-daily:
           release: xenial
           nose_args: nose_args
@@ -49,6 +50,7 @@
         - email-server-crew
         - archive-results
     builders:
+      - curtin-vmtest-sync-images
       - vmtest-daily:
           release: bionic
           nose_args: nose_args
--- a/curtin/jobs-vmtest.yaml
+++ b/curtin/jobs-vmtest.yaml
@@ -15,19 +15,6 @@
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
-- job:
-    name: curtin-vmtest-sync-images
-    node: torkoal
-    triggers:
-        - timed: "H 11,23 * * *"
-    builders:
-        - shell: |
-            #!/bin/bash -x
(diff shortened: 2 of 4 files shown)


curtin/jobs-vmtest-daily.yaml
```diff
@@ -30,6 +30,7 @@
         - email-server-crew
         - archive-results
     builders:
+      - curtin-vmtest-sync-images
```
OddBloke: Should this just be included as the first builder in the existing `vmtest-{daily,proposed}` builders?

OddBloke: Oh, and `curtin-vmtest`, I suppose.

curtin/jobs-vmtest.yaml
```diff
@@ -157,8 +149,32 @@
...
+            test "$(find /srv/images/.vmtest-data -maxdepth 0 -mmin -1080)" && exit 0
+
+            # Is vmtest-sync-images already running?
+            vsc_count=$(pgrep -c vmtest-sync-images)
+            rc=$?
+            ((rc > 1)) && exit 1 # pgrep error
+            ((vsc_count == 1)) && exit 0 # vmtest-sync-images already running
+            ((vsc_count >= 2)) && exit 1 # many vmtest-sync-images running => heads-up
+
+            git clone https://git.launchpad.net/curtin curtin-sync-images || exit 1
```
OddBloke: Would `set -e` above this line be a little cleaner? (I'm not sure myself.)

curtin/jobs-vmtest.yaml
```diff
@@ -157,8 +149,32 @@
...
+            set -ufx -o pipefail
+
+            # The images have been updated recently enough => exit 0
+            test "$(find /srv/images/.vmtest-data -maxdepth 0 -mmin -1080)" && exit 0
+
+            # Is vmtest-sync-images already running?
+            vsc_count=$(pgrep -c vmtest-sync-images)
+            rc=$?
+            ((rc > 1)) && exit 1 # pgrep error
+            ((vsc_count == 1)) && exit 0 # vmtest-sync-images already running
```
OddBloke: We don't need to wait here because the tests take care of blocking on the sync completing themselves, right?

curtin/jobs-vmtest.yaml
paride: Yes and no. Using `set -e` would allow us to drop some `|| exit 1`, but on the other hand would make other parts of the script less obvious, or at least more verbose. For example
```
test "$(find /srv/images/.vmtest-data -maxdepth 0 -mmin -1080)" && exit 0
```
```
vsc_count=$(pgrep -c vmtest-sync-images)
```
are expected to return nonzero, and that's fine, as it is expected and handled.

More in general I learned to use `set -e` less, mostly for the reasons explained here:

 - http://mywiki.wooledge.org/BashFAQ/105
 - https://hackmd.io/IoVN08RrQV6Hyb6ztY4kwg

My take is: while I mostly stopped liking it, I'm still using it for very simple scripts made by a few commands to be executed one after the other and expected to suceed, with no subshells or functions. The moment I have to *think* of that `set -e` actually does in order to write the script, I just drop it and handle the errors properly.

paride: Re-reading your comment, I see you were suggesting to add `set -e` *specifically* above that line. Then it would make sense, yes! But I try to avoid changing the shell options at mid-script, as where reading part of the script you never know if the code above changed the shell options. A subshell makes it clearer, e.g.
```bash
(
    set -e
    git clone https://git.launchpad.net/curtin curtin-sync-images
    cd curtin-sync-images
    ./tools/vmtest-sync-images
)
```
but honestly I don't think it's worth it.

OddBloke: I agree. Thanks for taking the time to consider it!

OddBloke:
+1, will leave for you to merge as there's a job deletion involved so you may need to hand-hold. Thanks!

PR #84, opened 2019-10-18
On failure, archive all the things

raharper:
When we have the failure to ssh issue, it's not possible to debug this from what's currently collected.  This is an attempt to archive all of the output on failure such that we can work out why things failed. 

Local results dir for a successful run come in at 2G per dir.  So this may be too much space.  Interested in discussing other ways to capture the qcow snapshot when we fail to ssh connect.

Also, not sure if this achieves my desired goal, so looking for feedback on change independent of whether it's a good idea long term.

### DIFF
--- a/cloud-init/default.yaml
+++ b/cloud-init/default.yaml
@@ -85,4 +85,3 @@
     publishers:
       - archive:
           artifacts: 'cloud-init/results/**'
-          excludes: 'cloud-init/results/*log, **/*img, **/*qcow2'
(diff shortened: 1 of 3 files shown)


powersj:
Hmm so this updated for the cloud-init autoland job, what job is actually failing? If it really is autoland, I think this would be ok as long as we don't keep too many of them.

raharper:
I don't want to change autoland;  so I put it in the wrong spot; I'm looking to ensure that the cloud-init ci jobs which run the integration tests can archive qcow2 files on failure of the cloud_tests run.

PR #85, opened 2019-10-24
curtin-ci: retry git clone and fetch a few times

paride:
This makes the job less prone to fail in case of network hiccups.

### DIFF
--- a/curtin/jobs-ci.yaml
+++ b/curtin/jobs-ci.yaml
@@ -105,17 +105,15 @@
           #!/bin/bash
           set -eux
 
+          # Retry a command 5 times over 30 minutes with cubic backoff
+          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
+
           branch=${landing_candidate//lp:/https://git.launchpad.net/}
-
-          cmd="git clone --branch=${landing_candidate_branch} $branch curtin-${BUILD_NUMBER}"
-          for i in 0s 5s 1m 10m; do sleep $i; $cmd && s=0 && break || s=$?; done; (exit $s)
-
+          retry git clone --branch=${landing_candidate_branch} $branch curtin-${BUILD_NUMBER}
           cd curtin-${BUILD_NUMBER}
           git rev-parse HEAD
           git remote add upstream https://git.launchpad.net/curtin
-
-          cmd="git fetch upstream --tags"
-          for i in 0s 5s 1m 10m; do sleep $i; $cmd && s=0 && break || s=$?; done; (exit $s)
+          retry git fetch upstream --tags
 
           no_proxy=launchpad.net https_proxy="http://squid.internal:3128" tox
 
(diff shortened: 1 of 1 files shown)


curtin/jobs-ci.yaml
```diff
@@ -106,11 +106,16 @@
           set -eux
 
           branch=${landing_candidate//lp:/https://git.launchpad.net/}
-          git clone --branch=${landing_candidate_branch} $branch curtin-${BUILD_NUMBER}
+
+          cmd="git clone --branch=${landing_candidate_branch} $branch curtin-${BUILD_NUMBER}"
+          for i in 0s 5s 1m 10m; do sleep $i; $cmd && s=0 && break || s=$?; done; (exit $s)
```
blackboxsw: do we want the exit $s  there? won't that also prevent us from getting to any steps below?

curtin/jobs-ci.yaml
```diff
@@ -106,11 +106,16 @@
...
+
           cd curtin-${BUILD_NUMBER}
           git rev-parse HEAD
           git remote add upstream https://git.launchpad.net/curtin
-          git fetch upstream --tags
+
+          cmd="git fetch upstream --tags"
+          for i in 0s 5s 1m 10m; do sleep $i; $cmd && s=0 && break || s=$?; done; (exit $s)
```
blackboxsw: Since the retry sleep counts are reused by both git commands, maybe it's worth either a function a variable

RETRY_SLEEPS=(0s 5s 1m 10m)

for i in ${RETRY_SLEEPS[@]}; do sleep $i; $cmd && break || s=$?; done;

curtin/jobs-ci.yaml
paride: The `(exit $s)` happens in a subshell, so it doesn't make the main script exit, and was needed to preserve the exit status of the command to retry (in case of failure). Now that I re-read it, it's probably not even necessary. However I changed my mind on how to do the retry, so this is probably not relevant anymore, see below :)

curtin/jobs-ci.yaml
paride: Thanks Chad; actually it's the rest of the retry logic that I would like not to repeat. I made it into a function, leaving the retry intervals customizable, but after thinking on it a bit more I'm convinced that we can have some default that will cover almost every case of our interest. So I propose this function:
```
retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
```
which has some nice properties:
 - Retries 5 times over half an hour (that is, has strong defaults)
 - It retries with a cubic backoff strategy, waiting little time after the first failures, increasing the wait time if the command's keeps failing
 - `"${@:1}"` expands as a bash array, which is way better than using a variable
 - Preserves the exit code
 - Runs in a subshell, thus not leaking the `for` index variable
 - It's short and reusable
 - It's shellcheck clean

paride:
Pushed a new commit with the new `retry()` function. Once the function is defined to retry a command it's sufficient to add a `retry` in front of it.

PR #87, opened 2019-10-29
metrics: retry the commands accessing the network a few times

paride:
The commands are retried 4 times over about 13 minutes with cubic backoff.

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -28,7 +28,7 @@
           retry git clone https://github.com/CanonicalLtd/metrics.git
           cd metrics
           export METRICS_PROMETHEUS=10.25.2.225:9091
-          retry python3 -m metrics.{metric_module}
+          python3 -m metrics.{metric_module}
 
 - builder:
     name: metric-influxdb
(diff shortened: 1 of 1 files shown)


paride:
@OddBloke this should save us from from those failure emails in case of temporary network issues. What do you think?

metrics/metrics.yaml
```diff
@@ -21,38 +21,47 @@
...
+
+          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
+
           rm -rf *
-          git clone https://github.com/CanonicalLtd/metrics.git
+          retry git clone https://github.com/CanonicalLtd/metrics.git
           cd metrics
           export METRICS_PROMETHEUS=10.25.2.225:9091
-          python3 -m metrics.{metric_module}
+          retry python3 -m metrics.{metric_module}
```
OddBloke: I'm not 100% sure that all the metrics modules will have been written to be retried this way. One scenario that occurs to me: if they perform multiple Prometheus inserts and we only start seeing failures after the first one, then we will retry and re-perform that first Prometheus insert. This _could_ result in some confusing and/or inconsistent InfluxDB data.

powersj: I agree to not retrying the actual running of the metrics. FWIW, we have entirely migrated off of Prometheus.

paride: Yeah I was not sure on retrying the DB inserts for this very reason. IIUC InfluxDB should not create duplicates if the same data point is inserted twice, so retrying the insert should not be different from re-running the Jeniks job, and thus should be safe.

However I'm not entirely sure either it is *actually* 100% safe, and as you say retrying the `git clones` should be enough to get rid of most of the failures, so ACK for dropping the retry on the metrics line.

PR #88, opened 2019-10-29
cloud-init integration-nocloud-kvm: retry to 'git clone' a few times

paride:
Retry the 'git clone' commands 5 times over half an hour with cubic
backoff. This should prevent temporary networking of Launchpad issues
from making the job fail.


paride:
Sometimes the `tox` run fails because it fails to pull the required dependencies, see for example:

https://jenkins.ubuntu.com/server/view/cloud-init,%20curtin,%20streams/job/cloud-init-integration-nocloud-kvm-d/215/

however I don't think we should retry it.

PR #89, opened 2019-10-30
curtin vmtest-devel: retry to git clone/fetch a few times

paride:
Workaround for transient network or Launchpad issues.


paride:
The change to `vmtest-devel` could have prevented this failure: https://jenkins.ubuntu.com/server/view/cloud-init,%20curtin,%20streams/job/curtin-vmtest-devel-arm64/40/.

The change to the autoland job makes it behave as the `curtin-ci` job, as it should be.

PR #91, opened 2019-11-06
mirror: git sync from github and mirror to launchpad

blackboxsw:
Currently jenkins mirrors cloud-init from launchpad's git repo into github.

This change has jenkins mirror FROM github TO launchpad.

I've tested each script line (with the exception of the final git push --mirror to launchpad) on our jenkins master.

Any gotchas we should be worried about?


raharper:
LGTM

PR #93, opened 2019-11-15
Refactor the vmtest parameters and add the curtin-cloudinit-sru job

paride:
Changes:
 - vmtest-add-repos: skip the intermediate vmtest_add_repos variable
 - Set the vmtest-add-repos parameter default from a variable
 - Add the curtin-cloudinit-sru job

### DIFF
--- a/curtin/jobs-vmtest.yaml
+++ b/curtin/jobs-vmtest.yaml
@@ -238,3 +238,34 @@
           [ $RET -eq 2 ] && RET=0
 
           exit $RET
+
+- job:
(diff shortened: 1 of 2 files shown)


paride:
It still surprises me that there is no way to tell JJB to always expand a macro, even if no parameter is passed. This would spare us from passing the empty `default_nose_args` and `default_vmtest_add_repos` in most of the job definitions using those parameters.

paride:
@raharper we're happy with this right?

PR #94, opened 2019-11-28
cloud-init-style-tip: export no_proxy=launchpad.net

paride:
Launchpad is not unreachable via squid.internal.


paride:
Fixes failures like: https://jenkins.ubuntu.com/server/job/cloud-init-style-tip/559/

Despite the move to GitHub/Travis I think this is a job worth keeping, what do you think?

PR #98, opened 2019-12-17
Output the command that's being retried

raharper:
The retry logic hid the exact command that's being retried.  Add
an echo in the loop so we can see exactly what is being run.

### DIFF
--- a/cloud-init/integration-nocloudkvm.yaml
+++ b/cloud-init/integration-nocloudkvm.yaml
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && echo "Trying: ${*:1}" >&2 && "${{@:1}}" && break; done)
+          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && echo "Trying: ${{*:1}}" >&2 && "${{@:1}}" && break; done)
 
           release="{release}"
           if [ -z "$TMPDIR" -a "$(hostname)" = "torkoal" ]; then
--- a/cloud-init/integration-nocloudkvm.yaml
+++ b/cloud-init/integration-nocloudkvm.yaml
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
+          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
 
           release="{release}"
           if [ -z "$TMPDIR" -a "$(hostname)" = "torkoal" ]; then
--- a/cloud-init/integration-nocloudkvm.yaml
+++ b/cloud-init/integration-nocloudkvm.yaml
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
+          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${{@:1}}"; "${{@:1}}";}} && break; done)
 
           release="{release}"
           if [ -z "$TMPDIR" -a "$(hostname)" = "torkoal" ]; then
--- a/cloud-init/integration-nocloudkvm.yaml
+++ b/cloud-init/integration-nocloudkvm.yaml
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${{@:1}}"; "${{@:1}}";}} && break; done)
+          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && echo "Trying: ${*:1}" >&2 && "${{@:1}}" && break; done)
 
           release="{release}"
           if [ -z "$TMPDIR" -a "$(hostname)" = "torkoal" ]; then
--- a/curtin/jobs-ci.yaml
+++ b/curtin/jobs-ci.yaml
@@ -106,7 +106,7 @@
           set -eux
 
           # Retry a command 5 times over 30 minutes with cubic backoff
-          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
+          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
 
           branch=${landing_candidate//lp:/https://git.launchpad.net/}
           retry git clone --branch=${landing_candidate_branch} $branch curtin-${BUILD_NUMBER}
@@ -163,7 +163,7 @@
           #!/bin/bash
           set -eux
 
-          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
+          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
 
           retry git clone lp:curtin
           cd curtin
--- a/curtin/jobs-ci.yaml
+++ b/curtin/jobs-ci.yaml
@@ -106,7 +106,7 @@
           set -eux
 
           # Retry a command 5 times over 30 minutes with cubic backoff
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${{@:1}}"; "${{@:1}}";}} && break; done)
+          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${@:1}"; "${@:1}";}} && break; done)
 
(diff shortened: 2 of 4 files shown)


cloud-init/integration-nocloudkvm.yaml
```diff
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
+          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
```
OddBloke: I believe the CI failures are happening because these curly braces are no longer escaped:

```suggestion
          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${{@:1}}"; "${{@:1}}";}} && break; done)
```

OddBloke: (I'll let you change the other instances of this, rather than proposing a suggestion for each one.)

raharper: Right, so all places using retry() need the double curly due to JJB expansion.

OddBloke:
On Tue, Dec 17, 2019 at 09:03:55AM -0800, Ryan Harper wrote:
> raharper commented on this pull request.
> 
> 
> 
> > @@ -80,7 +80,7 @@
>        - shell: |
>            #!/bin/bash
>  
> -          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
> +          retry() (for i in {0..5}; do sleep $((8*i**3)) && {echo "${@:1}"; "${@:1}";} && break; done)
> 
> Right, so all places using retry() need the double curly due to JJB expansion.

Annoying, not _all_.  If there are no parameters passed, then JJB
doesn't put the string through template expansion, so the double curly
braces end up present in the job XML.

I expect that retaining double braces where they already were (and not
adding them where they weren't) is the way to go.

raharper:
Well, they weren't consistent, specifically in retry() which runs the input args, skipping the first element.

"${@:1}"  

I see it like the above, and escaped:

"${{@:1}}"

So not clear to me what it should be... thoughts?

raharper:
```
% grep -nri "retry()" *
cloud-init/integration-nocloudkvm.yaml:83:          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
curtin/jobs-vmtest.yaml:202:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
curtin/jobs-ci.yaml:109:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
curtin/jobs-ci.yaml:166:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
metrics/metrics.yaml:25:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
metrics/metrics.yaml:40:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
metrics/metrics.yaml:57:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
metrics/metrics.yaml:83:          retry() (for i in {0..4}; do sleep $((8*i**3)) && "${@:1}" && break; done)

```

OddBloke:
On Tue, Dec 17, 2019 at 09:15:01AM -0800, Ryan Harper wrote:
> ```
> % grep -nri "retry()" *
> cloud-init/integration-nocloudkvm.yaml:83:          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
> curtin/jobs-vmtest.yaml:202:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
> curtin/jobs-ci.yaml:109:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
> curtin/jobs-ci.yaml:166:          retry() (for i in {0..5}; do sleep $((8*i**3)) && "${@:1}" && break; done)
> metrics/metrics.yaml:25:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
> metrics/metrics.yaml:40:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
> metrics/metrics.yaml:57:          retry() (for i in {{0..4}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
> metrics/metrics.yaml:83:          retry() (for i in {0..4}; do sleep $((8*i**3)) && "${@:1}" && break; done)
> 
> ```

The correct spelling depends, very annoyingly, on the context in which
it is used.  So the single-brace cases here aren't templated, and the
double-braced cases _are_.  Given that things are working as we expect
at the moment, you should just retain/match the current number of
braces in each place.

raharper:
OK.  I think that should match up.

paride:
One minute before merging, let me try one thing.

cloud-init/integration-nocloudkvm.yaml
```diff
@@ -80,7 +80,7 @@
       - shell: |
           #!/bin/bash
 
-          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && "${{@:1}}" && break; done)
+          retry() (for i in {{0..5}}; do sleep $((8*i**3)) && {{echo "${{@:1}}"; "${{@:1}}";}} && break; done)
```
paride: In the `{ }` compound command bash needs spaces separating the curly braces and the commands.

raharper: I thought the {{ was getting replaced with { by JJB?

raharper:
> Alternative implementation we may prefer:
> 
> ```
> retry() (for i in {0..4}; do sleep $((8*i**3)) && (set -x; "${@:1}") && break; done)
> ```
> 
> uses a subshell and `set -x` to print the command without resorting to `echo`, or:

I don't want set -x; I just want to see what command we're going to run in retry;

> 
> ```
> retry() (for i in {0..4}; do sleep $((8*i**3)) && { ((i)) && echo "Retrying ${@:1}"; "${@:1}"; } && break; done)
> ```
> 
> echoes only when retrying

If we succeed on the first try, then we'll never see the command.

cloud-init/integration-nocloudkvm.yaml
paride: It is, bash needs spaces around the curly braces. You can try in your terminal: `{echo foo;}` won't work, `{ echo foo; }` will work.

paride:
@raharper I'm fine with just echoing, the "request changes" is just for the missing spaces. But let me explain why I proposed the alternative implementations.

> > Alternative implementation we may prefer:
> > ```
> > retry() (for i in {0..4}; do sleep $((8*i**3)) && (set -x; "${@:1}") && break; done)
> > ```
> > 
> > 
> > uses a subshell and `set -x` to print the command without resorting to `echo`, or:
> 
> I don't want set -x; I just want to see what command we're going to run in retry;

What I don't like of echoing is that it's a bit weak. For example:



> 
> > ```
> > retry() (for i in {0..4}; do sleep $((8*i**3)) && { ((i)) && echo "Retrying ${@:1}"; "${@:1}"; } && break; done)
> > ```
> > 
> > 
> > echoes only when retrying
> 
> If we succeed on the first try, then we'll never see the command.

raharper:
> In the end, if we want `retry()` to be more verbose, what I'm for is:
> 
> ```
> retry() (for i in {0..4}; do sleep $((8*i**3)) && echo "Trying: ${*:1}" >&2 && "${@:1}" && break; done)
> ```
> 
> which is
> 
> * printing the "Trying" word makes it easier to understand where the printout comes from
> * the "Trying" words protects `echo` from things like `-e`
> * prints the error to stderr, so the output of a retried command can be parsed unmodified

I'm +1 for this, my primary ask was to just see what gets called the first time;

In curtin-vmtest-devel-debug, all we see in console output is:

```
[curtin-vmtest-devel-debug] $ /bin/bash /tmp/jenkins1484422546876077862.sh
test filter args: ""
Cloning into 'curtin-207'...

```
And I locally modified the bash script on the job so I could see what we actually cloned with this:

```
echo "Trying: git clone --branch=${landing_candidate_branch} ${landing_candidate} curtin-${BUILD_NUMBER}"
retry git clone --branch=${landing_candidate_branch} ${landing_candidate} curtin-${BUILD_NUMBER}
```

Which produced this output, which is all I wanted:
```
[curtin-vmtest-devel-debug] $ /bin/bash /tmp/jenkins4182383636444802716.sh
test filter args: ""
Trying: git clone --branch=fix/vmtest-bcache-missing-environ-vars https://git.launchpad.net/~raharper/curtin curtin-209
Cloning into 'curtin-209'...
```

and as I went to make the change, it made more sense to me to have retry emit what it's running.

As long as stderr is also captured in the console log, then I'll take your suggested change and apply it to this PR.

paride:
Yes it is captured by Jenkins.

PR #103, opened 2020-01-24
COPR: switch to python3 and replace run-centos with run-container

paride:
The run-centos script is deprecated in favor of the more general
run-container.

### DIFF
--- a/cloud-init/centos.yaml
+++ b/cloud-init/centos.yaml
@@ -35,11 +35,10 @@
           git clone --single-branch https://git.launchpad.net/cloud-init
           pushd cloud-init
 
-          export PYVER=python2
           export http_proxy="http://squid.internal:3128"
           export https_proxy="$http_proxy"
 
-          retry_cmd="./tools/run-centos 7 --srpm --artifact"
+          retry_cmd="./tools/run-container centos/7 --source-package --artifacts=. --pyexe=python3"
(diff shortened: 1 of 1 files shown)


cloud-init/centos.yaml
```diff
@@ -35,11 +35,10 @@
           git clone --single-branch https://git.launchpad.net/cloud-init
           pushd cloud-init
 
-          export PYVER=python2
           export http_proxy="http://squid.internal:3128"
           export https_proxy="$http_proxy"
 
-          retry_cmd="./tools/run-centos 7 --srpm --artifact"
+          retry_cmd="./tools/run-container centos/7 --source-package --artifacts=. --pyexe=python3"
```
OddBloke: Instead of passing in `pyexe` here, should we be updating `run-container` to default to Python 3 for everything? (And perhaps dropping Python selection support entirely, as Python 2 no longer works?)

paride: +1. I'll address this on the cloud-init side then, and then update this PR accordingly.

paride:
Relevant PR on the cloud-init side:

https://github.com/canonical/cloud-init/pull/192

powersj:
@paride your cloud-init PR was merged, but I think you still need to update this one, correct?

paride:
Superseded by #107.

PR #108, opened 2020-03-05
Run curtin-ci on amd64 only, curtin-autoland-test run on all the archs

paride:


powersj:
These basic tests do not take too long. Correct? As such I would prefer we run them on all architectures if we are capable of doing so at this point.

paride:
Per standup discussion it's probably a better idea to:

1. run the curtin-ci job on amd64 only
2. run the curtin-autoland-test job on *all* the archs

In this way we'll minimize the job runs on non-amd64 machines, and have one final run that tests everything on all the archs.

powersj:
ok +1

paride:
This will require deleting the jobs from jenkins, as there is a change in the job type (matrix vs. "normal").

PR #112, opened 2020-03-27
Triage email: list the top 5 subiquity crash reports per day

paride:
Plus minor, mainly cosmetic changes.


paride:
Today's second issue of the triage email has been generated from this branch, so you can see how it looks like.

The "Function" line is a bit long (= looks ugly), but I think it's useful to have it listed.

PR #113, opened 2020-03-31
tableflip-triage-email.sh: restore Reply-To: $recipients

OddBloke:
Otherwise replies go to the Jenkins email address, which can't receive
email.

### DIFF
--- a/daily-triage/tableflip-triage-email.sh
+++ b/daily-triage/tableflip-triage-email.sh
@@ -168,6 +168,7 @@ subject="Daily triage for: $projects [$triager]"
     cat <<-EOF
 	From: server@jenkins.canonical.com
 	To: $recipients
+	Reply-To: $recipients
 	Subject: $subject
 	MIME-Version: 1.0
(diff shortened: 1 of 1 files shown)


paride:
@OddBloke the script used to set the Reply-To header, but I removed it as now the email is sent to a proper mailing list, so it should be up to the mailing list manager (mailman3) to add the reply-to header, if set up to do so. I'm not even sure the header would be preserved after the email has been processed and re-sent by mailman. I think the right thing to do is to configure the mailing list to add the header. It was certainly possible with mailman2, I'm not used to mailman3 and I don't have admin access to our list's settings (@powersj has), but I expect the setting to be still there.

Thunderbird is smart, detects the email comes from a mailing list and adds a "Reply to list" button. I think it watches for the List-Id header or something. I think mutt has something similar: https://www.jefftk.com/p/list-reply-in-mutt.

OddBloke:
Aha, the mutt functionality is what I was missing, I think. Thanks!

PR #118, opened 2020-04-21
curtin-ci: keep 7 days of job results, up to 40 jobs 

paride:
Keep 10 curtin-ci builds, up from the 3 we are keeping by default. One
build is about 26MB big, so we can easily afford to keep 10.

### DIFF
--- a/curtin/jobs-ci.yaml
+++ b/curtin/jobs-ci.yaml
@@ -67,6 +67,10 @@
             - metal-ppc64el
             - metal-arm64
             - metal-s390x
+    properties:
+      - build-discarder:
+          days-to-keep: 7
+          num-to-keep: 40
     parameters:
       - landing-candidate
       - landing-candidate-branch


paride:
@OddBloke should address your comment from yesterday. I don't think we can push to the repo yet, at least I can't.

curtin/jobs-ci.yaml
```diff
@@ -67,6 +67,9 @@
             - metal-ppc64el
             - metal-arm64
             - metal-s390x
+    properties:
+      - build-discarder:
+          num-to-keep: 10
```
OddBloke, line 72: [`build-discarder`](https://docs.openstack.org/infra/jenkins-job-builder/properties.html#properties.build-discarder) has a `days-to-keep` parameter, should we use that so that even on heavy days we don't lose history?

Per the documentation of the configuration option (which I could only find on a job configuration page):

> [`days-to-keep` and `num-to-keep`] can be active at the same time, so you can keep builds for 14 days, but only up to a limit of 50 builds, for example. If either limit is exceeded, then any builds beyond that limit will be discarded. 

So we could bump this up to, say, 40 to be sure that we'll never have more than a gigabyte of jobs (in case we end up with the automation going wild and queuing a job every 15 minutes, for example), and then keep it lower than that in practice by keeping the results around for a week at most. What do you think?

paride: I think it's an excellent idea, I updated the branch.

PR #124, opened 2020-05-28
curtin: add vmtest jobs for proposed-focal and daily for groovy

blackboxsw:
With vmtests we want to make sure to avoid overloading our CI infrastructure, so cron jobs need to be scheduled on different days.

This schedule 'striping' is especially important for curtin vmtests because the load is so heavy and test runs are so long. 

- Add a vmtest groovy job for curtin to get once weekly CI coverage on the 6th day of the week.
- Adjust curtin-vmtest-daily-f dropping this 3rd test schedule from days 2,5,6 to just 2,5

- Add a proposed vmtest for focal (so that vmtests can run against what cloud-init has uploaded
   to the -proposed pocket for automated SRU verification into focal.

### DIFF
--- a/curtin/jobs-vmtest-proposed.yaml
+++ b/curtin/jobs-vmtest-proposed.yaml
@@ -59,6 +59,20 @@
           release: eoan
           triggerwhat: curtin-vmtest-proposed-e
 
+- job:
+    name: curtin-vmtest-proposed-f-trigger
+    node: torkoal
+    triggers:
+      - timed: "H 10 * * 3,6"
+    publishers:
+      - email-server-crew
+    wrappers:
+      - workspace-cleanup
+    builders:
+      - vmtest-proposed-trigger:
+          release: focal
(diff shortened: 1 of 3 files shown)


curtin/jobs-vmtest-proposed.yaml
```diff
@@ -59,6 +59,20 @@
...
+    triggers:
+      - timed: "H 10 * * 3,6"
+    publishers:
+      - email-server-crew
+    wrappers:
+      - workspace-cleanup
+    builders:
+      - vmtest-proposed-trigger:
+          release: focal
+          triggerwhat: curtin-vmtest-proposed-f
```
TheRealFalcon, line 74: Should there be a separate "curtin-vmtest-proposed-f" job?

blackboxsw: Thanks @TheRealFalcon for a good review. Correct, I worked too quickly and only defined the trigger and not the actual job. Just pushed changes.

PR #126, opened 2020-06-17
metrics: timestamp the builds and add a timeout

### LINKED ISSUE #125: Add timestamps to metric jobs
When dealing with an internal outage, I saw this in the log:

```
Trying: nc -zv <snip>
nc: connect to <snip> (tcp) failed: Connection refused
Trying: nc -zv <snip>
nc: connect to <snip> (tcp) failed: Connection refused
Trying: nc -zv <snip>
nc: connect to <snip> (tcp) failed: Connection refused
```

but I'm not able to discern how recently those errors happened.

It would be great if we could add timestamps to the job configurations, so that I can.

paride:
A timeout of 30 minutes should good to prevent stuck jobs from being
unnoticed without killing occasionally slow jobs.

Closes #125

### DIFF
--- a/metrics/default.yaml
+++ b/metrics/default.yaml
@@ -58,6 +58,11 @@
     node: metal-amd64
     triggers:
       - timed: "H * * * *"
+    wrappers:
+      - timestamps
+      - timeout:
+          timeout: 30
+          fail: true
     properties:
       - build-discarder:
           num-to-keep: 25


paride:
@OddBloke the `retry()` function used in the jobs can add up to ~13 minutes of sleep time, plus there's the time needed for the actual commands to run. I went for ~twice of that. Thanks for reviewing and merging!

PR #127, opened 2020-07-03
Bump the docker image size for the 'latest' image

paride:
Focal exceeds the previous 72MB limit.

### DIFF
--- a/docker/image-size.yaml
+++ b/docker/image-size.yaml
@@ -39,7 +39,7 @@
           description: "Ubuntu docker tag to pull."
       - string:
           name: MAXSIZE
-          default: !j2: "{{ {'latest': 72, 'rolling': 80, 'devel': 80}[dockertag] }}"
+          default: !j2: "{{ {'latest': 75, 'rolling': 80, 'devel': 80}[dockertag] }}"
           description: 'Maximum size of the Docker image'
     triggers:
       - timed: '@daily'


OddBloke:
Instead of bumping this, should we be pushing $someone to get it back under the limit?

paride:
Possibly! But then the question is: what should the limit be? The current limits where set by me when I first setup the jobs, adding a few MBs to what the current image sizes were. This was pre-Focal.

The idea of the job was to spot unreasonable changes in size, I don't think the change in size in this case is unreasonable; OTOH we all like small images :) This is not urgent, we can very well wait to hear what @mitechie thinks. I wouldn't know who to poke to make the images smaller.

OddBloke:
On Fri, Jul 03, 2020 at 06:23:28AM -0700, Paride Legovini wrote:
> Possibly! But then the question is: what should the limit be? The
> current limits where set by me when I first setup the jobs, adding a
> few MBs to what the current image sizes were. This was pre-Focal.

Yep, that's a good question to ask; I agree that waiting for Rick to
facilitate is the best path forward.

> The idea of the job was to spot unreasonable changes in size, I don't
> think the change in size in this case is unreasonable; OTOH we all
> like small images :)

The issue is that when we allow size to increase reasonably each
release, there can end up being an unreasonable jump in size between LTS
releases.

> This is not urgent, we can very well wait to hear what @mitechie
> thinks. I wouldn't know who to poke to make the images smaller.

CPC build the images, but it's likely that it's due to seeds/tasks so
Foundations are likely ultimately responsible.  But let's make that
Rick's problem. ;)

mitechie:
The agreement among the teams participating is that the images will naturally get slightly larger as new versions of software tend to have more code and lead to a slight increase each cycle/update. The goal of this alarm is to catch an accident such as the usrmerge change that recently brought in unexpected new dependencies that bloated the size of images. 

The standing rule of thumb is that if we identify the alarm goes off here that a brief investigation is done to look at what the delta is and if there are any surprises in the list of software included that was not expected. I can bring those to the sync calls with CPC and foundations we have on a regular basis and see if there's anything that we should be addressing. If it's a small change over time we can bump the limit and continue to look for the unexpected jumps. 

There's a different project to work on reducing our image sizes taking place but that is on a different track and is slightly orthogonal to the goal of this test. When that work is done we can look if we should reset the expected size to again be able to help catch an unexpected change.

mitechie:
Typing that out, I wonder if it might also be interesting or useful to dump out the package set and compare when that changes over time as part of the debug/tooling around this to make it easier to see what changes might have triggered the delta.

paride:
Good that you closed this as the Docker images are back into their original limits, with no need to bump them.

Logging the installed packages and their size should be a `dpkg-query -Wf '${Package} ${Installed-Size}\n' | column -t` away.

mitechie:
Nice I had heard the usrmerge fixes landed yesterday.

PR #128, opened 2020-07-03
cloud-init tip jobs: fix the tox environments

paride:
cloud-init commit a99c8bcea597c17d83d90bf1e09c9a5b68c3dc22 replaced the
tip-pycodestyle and tip-pyflakes tox environments with the tip-flake8
environment. This adapts the cloud-init-style-tip job to the change.


paride:
I think the `tip-pylint` run will fail, but for its own reasons, likely to be fixed in cloud-init.

PR #133, opened 2020-07-24
cloud-init: drop eoan jobs, add missing groovy jobs

paride:


cloud-init/integration-lxd.yaml
```diff
@@ -54,7 +41,7 @@
           threshold: FAILURE
     builders:
       - integration-lxd:
-          release: eoan
```
mitechie, line 57: should this go to groovy? It cannot yet because of lack of images or something? If we should move it again should we set a card reminder for a point in the future.

paride: From the diff it looks like I replaced `eoan` with `bionic`, but the `bionic` line will become part of the `cloud-init-integration-lxd-b` job defined above.

paride: The integration-lxd Groovy jobs are already present, here I just dropped Eoan.

PR #134, opened 2020-08-07
ubuntu-advantage-client: add jenkins job for ec2 instance/image cleanup

blackboxsw:
Jenkins to cleanup ec2 instances left around by ua-client CI.

Call server-test-scripts/ubuntu-advantage/ec2_cleanup.py to ensure all images vpcs, vms and keys are removed

Depends on https://github.com/canonical/server-test-scripts/pull/44 landing


ubuntu-advantage-client/+
```diff
@@ -0,0 +1,83 @@
+# Ubuntu Server QA Jenkins Jobs
```
TheRealFalcon, line 1: Is this file supposed to be here?

ubuntu-advantage-client/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+- project:
+    name: ubuntu-advantage-client
+    jobs:
+      - ubuntu-advantage-client-integration-ec2-terminate
+
+- defaults:
+    name: global
+    description: |
+      Cloud-init Ubuntu Server QA Jenkins Job
```
TheRealFalcon, line 26: Update description to UA

ubuntu-advantage-client/+
blackboxsw: oops failed :wq  dropped

PR #135, opened 2020-08-19
curtin: make curtainer use ubuntu-minimal images

paride:
This assumes the ubuntu-minimal-daily remote is correctly setup on the
Jenkins nodes. This is done by running:

  lxc remote add --protocol simplestreams ubuntu-minimal-daily \
      https://cloud-images.ubuntu.com/minimal/daily/

The jenkins-slave-setup Ansible playbook takes care of this, and also
sets up the ubuntu-minimal remote for released images.


mitechie:
Can you add the comments about how the images are added to the code maybe in case there's an error about it not being found the fix/place to look is right by the error message?

PR #138, opened 2020-08-28
Test the COPR build on CentOS 8

paride:
Test the copr build on centos 8 instead of centos 7. We are not even
building for centos 7 anymore due to missing Python 3 dependencies.

### DIFF
--- a/cloud-init/centos.yaml
+++ b/cloud-init/centos.yaml
@@ -23,7 +23,7 @@
     publishers:
       - email-server-crew
       - trigger:
-          project: cloud-init-copr-test-7
+          project: cloud-init-copr-test-8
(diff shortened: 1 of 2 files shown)


raharper:
What do you think about building 19.4-stable on centos7 ?

paride:
Should be easy, but I need to create a new project in copr that targets a different set of chroots. I'm adding a card so hopefully I won't forget to set it up.

paride:
Note that this change doesn't disable centos7 testing, as it's already not happening, at least not in a useful way. As we disabled building master in the epel-7 copr chroot this centos-7 job was testing the last rpm built on epel-7 again and again.

PR #139, opened 2020-09-16
Add pollinate test

paride:


pollinate/pollinate.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - shell: |
+          #!/bin/sh
+
+          set -eufx
+
+          # Discard the random data (stdout). The pollinate client
+          # normally prints useful logging messages to stderr, but
+          # only if started from a terminal. This means it will be
+          # quiet in Jenkins.
+          pollinate --testing --strict > /dev/null
```
panlinux, line 35: Thanks!
I looked at the job output, and have some comments:
- I'd like to see the URL it used, just so it's confirmed in the logs that it hit entropy.ubuntu.com
- if possible, I'd also like to have in the console output the version of the `pollinate` package that was used, and the ubuntu release it ran on

panlinux: Oh, if you keep the user-agent print that I saw in the job, then that satisfies the second bullet point above

panlinux (approved):
+1

PR #140, opened 2020-09-23
Add lma integration testing

paride:


lucaskanashiro:
I do not remember the last time I set up a Jenkins job but from what I can see it looks good and quite standard. Moreover, the output of the job seems satisfactory.

+1 for these changes.

PR #141, opened 2020-10-07
Add copr build of stable-19.4 on centos/7

paride:


cloud-init/centos.yaml
```diff
@@ -46,7 +46,41 @@
...
+      - timed: "@daily"
+    publishers:
+      - email-server-crew
+      - trigger:
+          project: cloud-init-copr-test-7
+          threshold: SUCCESS
+    wrappers:
+      - workspace-cleanup
+    builders:
+      - shell: |
```
OddBloke, line 65: This definition is (to my eyes) identical to the `cloud-init-copr-build` shell script, except `s/8/7/` in some places and `s/-dev/-dev-19.4/` in another. Is it worth moving them to a separate script file or builder definition, so we don't have to manually keep them in sync?

cloud-init/centos.yaml
```diff
@@ -65,3 +99,20 @@
...
+          #!/bin/bash
+          set -e
+
+          export http_proxy="http://squid.internal:3128"
+          export https_proxy="$http_proxy"
+
+          git clone --depth 1 https://github.com/canonical/server-test-scripts
+          cd server-test-scripts/cloud-init
+          ./copr_test -vv 7
```
OddBloke, line 118: Same Q here, but even moreso. ;)

paride:
I hit a limitation/bug of bionic's JJB while refactoring this: variable expansion in the form
```
{somevariable|default-value}
```
do not work when there is a dash in the default value side (so the above example wouldn't work). This is fixed in newer JJB versions. I'm preparing a PR to make the deploy job use it.

paride:
Here: https://github.com/canonical/server-jenkins-jobs/pull/142/

paride:
https://github.com/canonical/server-jenkins-jobs/pull/142 landed, this should be good for review/landing too.

PR #142, opened 2020-10-08
admin-jobs-update: pull the latest JJB in a venv

paride:


paride:
I wonder how solid the venv will be. This approach may come helpful in other jobs too, and `admin-jobs-update` can serve as a good testbed for it as it runs frequently and nothing super bad happens if it fails.

paride:
The plan is to reuse the venv handling snippet in other jobs needing packages from pip (e.g. the copr builds and the boot-speed jobs).

admin/deploy.yaml
```diff
@@ -30,10 +30,26 @@
...
-          cd server-jenkins-jobs
+          #!/bin/bash
+
+          set -ux -o pipefail
+
+          # Setup a venv and install JJB, reuse the venv if already existing.
+          # Avoid leaving a broken venv behind is something goes wrong here.
+          trap 'rm -rf job-venv' EXIT
+          [[ -d job-venv ]] || python3 -m venv job-venv || exit 1
+          source job-venv/bin/activate
```
OddBloke, line 41:
```suggestion
          . job-venv/bin/activate
```

`source` is an easily-avoidable bashism.

OddBloke (approved):
LGTM, thanks!

paride:
@OddBloke thanks! I'm fine with the `source` -> `.` change, however there are other bashisms in the script, eg `[[ ... ]]` tests.

I used to try to stick with posix sh scripts, but lately I'm basically doing the opposite: I use bash without being shy with bashisms unless I have a specific reason to use plan sh (speed or minimal environments). There are a lot of cases where bash allows to write clearer and more robust scripts.

OddBloke:
On Thu, Oct 08, 2020 at 08:16:27AM -0700, Paride Legovini wrote:
> @OddBloke thanks! I'm fine with the `source` -> `.` change, however
> there are other bashisms in the script, eg `[[ ... ]]` tests.

Yep, it's just a minor nit.

> I used to try to stick with posix sh scripts, but lately I'm basically
> doing the opposite: I use bash without being shy with bashisms unless
> I have a specific reason to use plan sh (speed or minimal
> environments). There are a lot of cases where bash allows to write
> clearer and more robust scripts.

My general feeling is that if I'm reaching for bash because POSIX
shell isn't sufficient then I should perhaps reach further and use
Python etc.  Jenkins scripts are a very reasonable exception to this
rule, though.

PR #143, opened 2020-10-13
Update the tableflip triage email

paride:


daily-triage/tableflip-triage-email.sh
```diff
@@ -18,12 +18,12 @@ command -v ubuntu-bug-triage
 
 
 # Projects to triage
-projects="cloud-init cloud-utils curtin simplestreams"
+projects="cloud-init cloud-utils simplestreams"
```
blackboxsw, line 21: Is it worth adding pycloublib?

paride:
Added pycloudlib, which also required adding support for projects which are only on github and not on LP.

This is still not perfect, as the script doesn't collect GitHub issues as it does for LP bugs. This can certainly be done, but first I'd like to experiment a bit with syncing GitHub PRs/issues with Trello. If it works for us, then we may want to *remove* things from the triage email instead of adding more stuff. In general the current situation is not ideal, as we are tracking the same stuff in multiple places (GitHub itself, Trello, the triage email). I'd like to at least try to make things converge to Trello a bit more.

daily-triage/tableflip-triage-email.sh
```diff
@@ -155,13 +160,18 @@ subject="Daily triage for: $projects [$triager]"
...
+            rm -f "$project-reviews.html"
         fi
     done
 
-    echo "<h5>Yesterday's top 5 subiquity crashes</h5>"
-    echo "<ul>"
-    cat subiquity_crashes.html
-    echo "</ul>"
+    for github_project in "${github_projects[@]}"; do
+        if [ -e "$github_project-reviews.html" ]; then
```
blackboxsw, line 168: This now exists always (and contains "None")  to we want to check that it is not None? Befor emitting the specific github project section at all should we be checking that the file doesn't contain None or do we always want this section to print.
```suggestion
        if [ "None" != "$( cat $github_project-reviews.html)" ]; then
```

blackboxsw (approved):
Thanks @paride, I'm good with this whatever your decision on redacting "None" github project sections or keeping them.

As for tools that might help sync github to trello:
   Maybe `apt-get install ghi`,or snap install hub  plus  `pip3 install py-trello` to  obtain open github issues and py-trello to push them into our board with an Oauth token.

daily-triage/tableflip-triage-email.sh
paride: I think a better solution is to delete the file if there are no reviews and leave the `[ -e` test here (and in the code that generated the plaintext email).

paride:
> As for tools that might help sync github to trello:
> Maybe `apt-get install ghi`,or snap install hub plus `pip3 install py-trello` to obtain open github issues and py-trello to push them into our board with an Oauth token.

Thanks for the pointers! I was thinking to use Robie's report2trello [1], which uses https://github.com/tghw/trello-py/ to talk to Trello. The one you suggested (py-trello) seems to be way more popular.

[1] https://git.launchpad.net/~racb/+git/report2trello

PR #144, opened 2020-10-14
Add vm jobs to ubuntu-advantange-client

lucasmoura:
Running the full set of uaclient test in travis CI is taking a lot of time (around 2h30), we want to move some of these tests to a daily jenkins job.

We are starting by moving the vm tests, which are currently not been run on travis due to time concerns as well. We will move other test in the near future.

PS: We cannot merge this branch until this [PR](https://github.com/canonical/ubuntu-advantage-client/pull/1204) lands in uaclient, since it sets some necessary definitions for vm tests to run

### DIFF
--- a/ubuntu-advantage-client/default.yaml
+++ b/ubuntu-advantage-client/default.yaml
@@ -17,7 +17,12 @@
 
 - project:
     name: ubuntu-advantage-client
+    ubunturelease:
+      - 16.04
+      - 18.04
+      - 20.04
     jobs:
+      - uaclient-vm-{ubunturelease}
       - ubuntu-advantage-client-integration-ec2-terminate
 
 - defaults:
--- a/ubuntu-advantage-client/jobs.yaml
+++ b/ubuntu-advantage-client/jobs.yaml
@@ -1,5 +1,5 @@
 # Ubuntu Server QA Jenkins Jobs
-# Copyright (C) 2016 Canonical Ltd.
+# Copyright (C) 2020 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -15,6 +15,28 @@
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
+- job-template:
+    name: uaclient-vm-{ubunturelease}
+    parameters:
+      - uaclient
+    wrappers:
+      - workspace-cleanup
+      - timeout:
+          timeout: 50
+          fail: true
+    publishers:
+      - email-server-crew
+    triggers:
+      - timed: '@daily'
+    builders:
+      - shell: |
+          #!/bin/bash
+          set -ex
+
+          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
+          cd ua-client
+          tox -e behave-vm-{ubunturelease}
+
 - job:
     name: ubuntu-advantage-client-integration-ec2-terminate
     node: torkoal
@@ -22,6 +44,8 @@
       - workspace-cleanup
     publishers:
       - email-server-crew
+    triggers:
+      - timed: '@daily'
     builders:
       - shell: |
           OLDER_THAN_DATE=`date -d yesterday +%m/%d/%y`
--- a/ubuntu-advantage-client/parameters.yaml
+++ b/ubuntu-advantage-client/parameters.yaml
@@ -1,0 +1,38 @@
+# Ubuntu Server QA Jenkins Jobs
+# Copyright (C) 2016 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version..
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+
+- parameter:
+    name: git-clone-params
+    parameters:
+      - string:
+            name: '{gitvarprefix}REPO'
+            default: '{giturl}'
+            description: '{gitproject} repository location'
+      - string:
+            name: '{gitvarprefix}BRANCH'
+            default: '{gitbranch|master}'
+            description: '{gitproject} branch to pull'
+
+
+# Sets the UAREPO and UABRANCH environment variables
+- parameter:
+    name: uaclient
+    parameters:
+      - git-clone-params:
+          gitproject: ubuntu-advantage-client
+          giturl: 'https://github.com/canonical/ubuntu-advantage-client.git'
+          gitvarprefix: UA


ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,6 +15,36 @@
...
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
+- job:
+    name: uaclient-vm-16.04
+    parameters:
+      - uaclient
+    builders:
+      - shell: |
+          #!/bin/bash -x
+          set -ex
```
paride, line 25: Duplicate `-x`.

ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,6 +15,36 @@
...
+
+          git clone ${uaclient} /tmp/uac
+          cd /tmp/uac
```
paride, line 28: I'd prefer the job to run in the Jenkins WORKSPACE instead of changing dir to `/tmp`, so everything is contained there. As is this will also hit "directory already existing" issues when re-running the job, as `/tmp/uac` is not cleaned up.

ubuntu-advantage-client/parameters.yaml
```diff
@@ -0,0 +1,24 @@
...
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+
+- parameter:
+    name: uaclient
+    parameters:
+      - string:
+          name: uaclient
+          default: https://github.com/canonical/ubuntu-advantage-client.git
+          description: git url to clone project from
```
paride, line 24: It is sometimes useful to be able to specify a branch to clone, so the tests can be easily run against a PR branch. I have these "template" parameters for doing this: https://github.com/canonical/server-jenkins-jobs/blob/master/lma/parameters.yaml

ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,6 +15,36 @@
...
+
+          TOXENV=behave-16-04
```
paride, line 30: Looking at https://github.com/canonical/ubuntu-advantage-client/pull/1204 is seems that the tox environments are called `behave-vm-16.04` (and similar). The TOXENV should probably be updated here.

ubuntu-advantage-client/default.yaml
```diff
@@ -18,6 +18,8 @@
 - project:
     name: ubuntu-advantage-client
     jobs:
+      - uaclient-vm-16.04
+      - uaclient-vm-18.04
```
paride, line 22: We should really use a job template here, something like:
```
ubunturelease:
  - 16.04
  - 18.04
  - 20.04
jobs:
  - uaclient-vm-{ubunturelease}
```

paride (changes requested):
Thanks for this! There are some things to adjust, but nothing critical. We may also want to add two wrappers to the jobs: `workspace-cleanup` and `timeout`. I'd also go for a shallow git clone unless the full git history is needed for some reason (e.g. past git tags).

I'll be happy to do the changes, I setup similar jobs several times now so I have all the relevant bits handy. OTOH if you want to get to know our Jenkins and JJB things better I'll also be happy to do some review back and forth.

paride:
The jobs call `./tools/run_lxd_integration_tests.sh`, but the script doesn't exist in the uac repository and isn't part of https://github.com/canonical/ubuntu-advantage-client/pull/1204/. Do you plan to add it in a separate PR?

Can we replace it with a direct `tox` call?

paride:
Added [WIP] in title mostly to signal this is blocked by https://github.com/canonical/ubuntu-advantage-client/pull/1204

lucasmoura:
@paride Thanks for the review. I have updated the code with the points you have raises. However, I am just unsure about the `ubunturelease` part. By just adding that into the yaml, the jobs will loop on them to construct the list of jobs to run ?

Also, thanks for spotting the typos on the tools script. Right now it is not a tox command because it need to perform additional steps before running the behave tests.

paride:
> However, I am just unsure about the `ubunturelease` part. By just adding that into the yaml, the jobs will loop on them to construct the list of jobs to run ?

Yes exactly. We need something like this in the project definition:
```
- project:
    name: ubuntu-advantage-client
    ubunturelease:
      - 16.04
      - 18.04
      - 20.04
    jobs:
      - uaclient-vm-{ubunturelease}
      - ubuntu-advantage-client-integration-ec2-terminate
```

The the `job` definition has to be changed to a `job-template` like:

```
- job-template:
    name: uaclient-vm-{ubunturelease}
    parameters:
      - uaclient
    wrappers:
      - workspace-cleanup
      - timeout:
          timeout: 50
          fail: true
    builders:
      - shell: |
          #!/bin/bash
          set -ex
          export TOXENV=behave-vm-{ubunturelease}

          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
          cd ua-client
          ./tools/run_lxd_travis_integration_tests.sh
```
Note that I used `{ubunturelease}` in the builder definition. The TOXENV variable will be set to `behave-vm-16.04` (and similar), which I think matches the actual environment names in https://github.com/canonical/ubuntu-advantage-client/pull/1204/.

ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,6 +15,39 @@
...
+          timeout: 50
+          fail: true
+    builders:
+      - shell: |
+          #!/bin/bash -x
+          set -e
+          export TOXENV=behave-vm-18-04
+
+          git clone --depth 1 -b "$STSBRANCH" "$STSREPO" sts
+          ./tools/run_lxd_travis_integration_tests.sh
```
paride, line 49: Let's also add a publisher here:
```
    publishers:
      - email-server-crew
```
so we'll get emailsl when the job build fails. (We actually don't because the `server-crew-qa` mailing list is broken, but we will once it's fixed.)

ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,6 +15,39 @@
...
+      - uaclient
+    wrappers:
+      - workspace-cleanup
+      - timeout:
+          timeout: 50
+          fail: true
+    builders:
+      - shell: |
+          #!/bin/bash -x
+          set -e
```
paride, line 45: Nit: let's have all the flags in one place (`set -ex`) instead of setting `-x` in the shebang and `-e` with `set`.

Or did you do it on purpose to also log the `set -e` bit?

paride (changes requested):

ubuntu-advantage-client/jobs.yaml
lucasmoura: No, there is no problem changing that at all

lucasmoura:
Thanks for the review @paride. I have updated the code

paride:
@lucasmoura I didn't test it, but this now LGTM, thanks! For some reason Travis is taking a long time to run.

Once https://github.com/canonical/ubuntu-advantage-client/pull/1204 lands I'll test and land this, let's keep it in WIP state for the moment, if you agree.

lucasmoura:
@paride The vm tests PR landed on uaclient. I think we can merge this one now

paride:
I deployed the jobs without merging to test them, but they're failing like this:

```
+ TOXENV=behave-vm-18.04
+ git clone --depth 1 -b master https://github.com/canonical/ubuntu-advantage-client.git ua-client
Cloning into 'ua-client'...
+ cd ua-client
+ ./tools/run_lxd_travis_integration_tests.sh
cp: cannot stat '*-debs/*': No such file or directory
```

If I don't go wrong that's because the job doesn't perform the .deb package building, so there are no `deb-artifacts` to archive, so the script will require some tweaking.

Anyway I had a better look at the `run_lxd_travis_integration_tests.sh` wrapper, and apparently what it's doing is mainly setting up lxd from snap and cleaning up the source tree, e.g. by moving the .debs away (the failing step above).

In general I prefer the Jenkins job not do perform `sudo` operation or to reconfigure services like `lxd` which are shared with other jobs. Remember the the jobs do not run in an ephemeral environment.

The assumption that the Jenkins nodes have snap LXD configured is safe, and indeed just running
```
tox -e behave-vm-{ubunturelease}
```
in the job's script seems to work, I did it manually here: https://jenkins.ubuntu.com/server/job/uaclient-vm-16.04/2/.

It seems to me that we could safely skip the wrapper scripts and just have tox run. What do you think?

lucasmoura:
@paride Thanks for that context. Yes, I agree that we can move the script to call tox directly. I will update the code

lucasmoura:
@paride I have updated the code

paride:
@lucasmoura thanks! Just another thing: the job definition doesn't setup a trigger, which means that the jobs will only run if started manually. I think we want to automatically run the jobs daily. In this case you should add a `timed` trigger to the `job-template`:
```
    triggers:
      - timed: '@daily'
```

paride (approved):

PR #146, opened 2020-11-05
Add Hirsute and template the -proposed jobs

paride:


paride:
The non -proposed jobs are "chained", one triggers the following, so they can't be templated so easily. Moreover each jobs calls the `ec2-template` job, which we may want to run independently and on a daily schedule.

In other words there's still room for improvement, but let's start with the -proposed jobs.

The -proposed jobs only run if there's actually a package in -proposed, so we can have them run for all the supported releases. The other jobs only run on the supported LTS releases + the devel release. I think it's enough.

The names of the proposed jobs will change (full release name instead of initial). This simplifies the job management quite a bit.

cloud-init/integration-proposed.yaml
```diff
@@ -15,215 +15,44 @@
...
+      - email-server-crew
+      - archive-results
 
 - builder:
     name: proposed-integration
     builders:
       - shell: |
           release="{release}"
-          release_ver="{release-ver}"
+          release_ver=$(ubuntu-distro-info --release --series={release} | cut -f ' ' -f 1)
```
paride, line 52: Pity that `ubuntu-distro-info` doesn't have a `--numeric-only` flag

OddBloke (approved):
This LGTM, thanks! I've reviewed both this diff, and a diff between the generated jobs (at https://paste.ubuntu.com/p/PHc7fd6SCJ/, for reference) and both look good.

paride:
Waiting to get feedback on https://github.com/canonical/server-jenkins-jobs/pull/147 before merging.

PR #147, opened 2020-11-05
admin-jobs-update: support passing --delete-all to JJB

paride:
Add a parameter to optionally make JJB delete the jobs it used to manage
but doesn't manage anymore (obsolete jobs, see [1]). Useful when Ubuntu
releases reach EOL and several jobs are to be removed at the same time.

Also: make JJB run on all the jobs directories even if it fails running
in one directory. In this case report failure on exit.

[1] https://docs.openstack.org/infra/jenkins-job-builder/execution.html#deleting-jobs-views

### DIFF
--- a/admin/deploy.yaml
+++ b/admin/deploy.yaml
@@ -23,7 +23,11 @@
     parameters:
       - bool:
           name: force_update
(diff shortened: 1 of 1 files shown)


paride:
PR updated separating the JJB flags from the JJB *command* flags

PR #148, opened 2020-11-05
oci unit: add postgresql

paride:


paride:
Note: I did some renaming, we'll need to delete the existing `lma-integration` job if you agree on the changes. Rationale in the commit message.

The postgresql job is defined with a project + job-template, so it will be very easy to add more jobs.

oci/integration.yaml
```diff
@@ -16,7 +16,7 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
 - job:
-    name: lma-integration
+    name: oci-integration-lma
```
sergiodj, line 19: You can drop the `-lma` suffix here, since this will cover more than the LMA packages.

sergiodj (changes requested):
Everything looks good!  The only comment I have is related to the `-lma` suffix, which is not needed anymore since we will be using this to test all OCI images, not only the LMA ones.

sergiodj:
> Note: I did some renaming, we'll need to delete the existing `lma-integration` job if you agree on the changes. Rationale in the commit message.

Yeah, that is totally fine.

oci/integration.yaml
paride: @sergiodj wouldn't it be better to have different jobs for different stacks? I was thinking of `oci-integration-lma` for LMA, `oci-integration-otherstack` for otherstack, and so on.

paride: Marking resolved per MM discussion.

PR #154, opened 2020-11-23
Add tox and a GitHub Action running it

paride:


paride:
The GH Action runs *way* faster than Travis. I know we could switch to .com, but the job is so simple that setting up the action is faster than requesting the switch, and we have other projects using actions already (e.g. cloud-init cla check, pycloudlib ci).

tox.ini
```diff
@@ -0,0 +1,14 @@
+[tox]
+envlist = yamllint, jjb-test
+skipsdist = true
+
+[testenv:yamllint]
+deps = yamllint==1.25.0
+commands = yamllint --strict .
+
+[jjb-base]
+deps = jenkins-job-builder==3.2.0
```
paride, line 10: I plan to add a `jjb-lint` environment for jenkins-job-linter. This will make more sense then.

OddBloke (approved):
LGTM, and I see the successful run on your personal repo.

paride:
I was (and still am) a bit puzzled by the fact that the action didn't run on the PR itself, as you noticed as you had to look for the results in my fork of the repo. The `pull_request` event runs on the *merge* of the PR itself and its target branch, and as the PR introduced the action I would expected to see it running.

OddBloke:
On Mon, Nov 23, 2020 at 08:48:28AM -0800, Paride Legovini wrote:
> I was (and still am) a bit puzzled by the fact that the action didn't
> run on the PR itself, as you noticed as you had to look for the
> results in my fork of the repo. The `pull_request` event runs on the
> *merge* of the PR itself and its target branch, and as the PR
> introduced the action I would expected to see it running.

I believe this is a security feature, but I don't recall the specifics.

PR #158, opened 2020-11-25
lma-integration: do not run tests with lxd-runner

paride:


paride:
The Jenkins job failed because LXD was stuck on the machine where the job ran. I've got it unstuck, but still we can skip `lxd-runner` to make things simpler and hopefully more reliable.

sergiodj:
LGTM, thanks!

sergiodj:
Something I remembered now: the output of `./run-tests` is different now because of the containerization work I did.  I'm assuming that there is nothing else that needs to be adjusted because of this.  Am I correct?  :-)

paride:
@sergiodj Jenkins cares about the exit code, and I don't think we're parting the output anywhere...

sergiodj:
@paride OK, that's what I thought.  Cool!

PR #160, opened 2020-12-08
[WIP] Add remaining uaclient jobs

lucasmoura:


jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,8 +15,307 @@
...
+              credential-id: ua-contract-token-staging
+              variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
+          - text:
+              credential-id: ua-azure-client-id
+              variable: UACLIENT_BEHAVE_AZ_CLIENT_ID
+          - text:
+              credential-id: ua-azure-client-secret
+              variable: UACLIENT_BEHAVE_AZ_CLIENT_SECRET
+          - text:
+              credential-id: ua-azure-tenant-id
```
blackboxsw, line 227: This is named ua-azure-tenant otherwise we get failures like https://jenkins.ubuntu.com/server/view/ua/job/uaclient-cleanup-stale-azure-vms/1/console

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,8 +15,307 @@
...
+          fail: true
+      - credentials-binding:
+          - text:
+              credential-id: ua-contract-token
+              variable: UACLIENT_BEHAVE_CONTRACT_TOKEN
+          - text:
+              credential-id: ua-contract-token-staging
+              variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
+          - text:
+              credential-id: ua-aws-access-key
```
blackboxsw, line 301: s/ua-aws-access-key/ua-aws-access-key-id/

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,8 +15,307 @@
...
+
+          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
+          cd ua-client
+
+          PACKAGE_BUILD_SERIES={ubuntureleasename} make travis-deb-install
+          PACKAGE_BUILD_SERIES={ubuntureleasename} make travis-deb-script
+
+
+- job-template:
+    name: uaclient-vm-{ubuntuvmrelease}
```
blackboxsw, line 148: While it's nice to have a simple job template that can be used for all uaclient-vm-{ubuntuvmrelease} release items, can we somehow ensure a time splay (distrubution) or a trigger sequence for these to make sure we aren't using KVM backend at the same time as other uaclient-vm jobs, cloud-init vm-based jobs and curtin vm-based jobs. Having too many kvm jobs running tends to kill our jenkins server.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,8 +15,307 @@
...
+      - shell: |
+          #!/bin/bash
+          set -ex
+
+          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
+          cd ua-client
+          tox -e behave-azurepro-{ubunturelease}
+
+- job-template:
+    name: uaclient-azure-generic-{ubunturelease}
```
blackboxsw, line 244: like the vm-based tests, I think we need to run azure-based tests sequentially and not in parallel especially because we will have to worry about hitting resource caps etc which had affected us in travis runs.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
paride: This is true, OTOH it's really nice to have a single job template. The better way to fix the concurrency problem is to use this plugin:

https://plugins.jenkins.io/throttle-concurrents/

which allows to define job classes and throttle the number of concurrent builds per class. For example we could define a `ua-kvm-job` class and restrict it to 1 job/node, or even better define a more general `heavy-vm-job`, to be used across projects (ua/curtin/...). This way we could limit heave VM jobs to (say) 4 concurrent jobs per node, and that's it, no need to "chain" jobs like in curtin. I'd like to make this change in curtin too.

Unfortunately we don't have the plugin in our Jenkins. There's a RT open for it (@mitechie knows already), but IS didn't get to it yet. I can try to push a bit more for it, but it's difficult to get an ETA.

The other possibility I see is to use a matrix job, and have `{ubunturelease}` as an axis. I think the list can still be defined at the project level. Let's speak at standup, I can put together a prototype job and see if it works for us. Otherwise we'll have to "chain" the jobs.

jenkins-jobs/ubuntu-advantage-client/default.yaml
```diff
@@ -17,12 +17,38 @@
...
+      - focal
+    ubuntuupgrade:
+      - 14.04
+      - 16.04
+      - 18.04
     jobs:
-      - uaclient-vm-{ubunturelease}
+      - uaclient-flake8
+      - uaclient-mypy
+      - uaclient-unittest
```
paride, line 41: While they do no harm, there's no need to have these jobs here, as they're not templated and don't use variables defined by the project. It's probably a bit safer not to have them here so they don't inherit variables we don't use.

jenkins-jobs/ubuntu-advantage-client/default.yaml
```diff
@@ -17,12 +17,38 @@
...
+      - uaclient-cleanup-stale-aws-vms
+      - uaclient-cleanup-stale-azure-vms
+      - uaclient-aws-pro-{ubunturelease}
+      - uaclient-azure-pro-{ubunturelease}
+      - uaclient-aws-generic-{ubunturelease}
+      - uaclient-azure-generic-{ubunturelease}
+      - uaclient-lxd-{ubunturelease}
+      - uaclient-vm-{ubuntuvmrelease}
+      - uaclient-upgrade-{ubuntuupgrade}
       - ubuntu-advantage-client-integration-ec2-terminate
```
paride, line 52: same for this one

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -15,8 +15,307 @@
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 
+- job:
+    name: uaclient-flake8
+    defaults: ubuntu-advantage-client
+    node: torkoal
```
paride, line 21: No need to limit these jobs to torkoal I think.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
paride: +1, see my other comment

paride (changes requested):

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
blackboxsw: Good suggestions @paride, as we mentioned in standup, let's go for matrix job at the moment to avoid too many concurrent vm-based jobs. And we can improve this when the plugin situation is sorted.

blackboxsw (approved):
LGTM changed azure-pro and  azure-generic to matrix and offset their timed job runs.

PR #161, opened 2020-12-09
curtin: send failure emails to mwhudson

paride:


paride:
@mwhudson this should let you get the jenkins failure emails.

@blackboxsw mentioned a mailing list that's not working. That's `server-crew-qa@lists.c.c`, we're waiting for IS to have a look. In any case it's better to add you as a recipient as I did here rather than subscribing you to the list. This way you'll only get the curtin emails (I doubt you're interested in the others!).

mwhudson:
Thanks for taking care of this!

PR #165, opened 2020-12-15
Add remaining jobs for uaclient

lucasmoura:
Add missing trusty flake8 and pytest jobs and fix xenial flake8 and pytest jobs to run with the correct python version


paride (approved):
Thanks!

PR #166, opened 2021-01-05
Add job to cleanup stale lxd vms

lucasmoura:
If the uaclient jobs fail to execute, they may leave
stale lxd vms that do not need to persist in the node.
We are now adding a script to cleanup those stale vms

### DIFF
--- a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
+++ b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
@@ -131,7 +131,7 @@
     wrappers:
       - workspace-cleanup
     triggers:
-      - timed: '@daily'
+      - timed: 'H 4 * * *'
     builders:
       - shell: |
           set -ex
@@ -172,7 +172,7 @@
               credential-id: ua-azure-subscription-id
               variable: UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID
     triggers:
-      - timed: 'H 1 * * *'
+      - timed: 'H 0,4 * * *'
     builders:
       - shell: |
           set -ex
@@ -191,6 +191,21 @@
           python server-test-scripts/ubuntu-advantage-client/azure_cleanup.py --client-id ${UACLIENT_BEHAVE_AZ_CLIENT_ID} \
           --client-secret ${UACLIENT_BEHAVE_AZ_CLIENT_SECRET} --subscription-id ${UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID} \
             --tenant-id ${UACLIENT_BEHAVE_AZ_TENANT_ID} || true
+
+- job:
+    name: uaclient-cleanup-stale-lxd-vms
+    defaults: ubuntu-advantage-client
+    node: torkoal
+    wrappers:
+      - workspace-cleanup
+    triggers:
+      - timed: 'H 1 * * *'
+    builders:
+      - shell: |
+          set -ex
+
+          git clone https://github.com/canonical/server-test-scripts.git
+          python3 server-test-scripts/ubuntu-advantage-client/lxd_cleanup.py --prefix ubuntu-behave
 
 - job-template:
     name: uaclient-build-package-{ubuntureleasename}
@@ -249,7 +264,7 @@
               credential-id: ua-contract-token-staging
               variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
     triggers:
-      - timed: '@daily'
+      - timed: 'H 2 * * *'
     builders:
       - shell: |
           #!/bin/bash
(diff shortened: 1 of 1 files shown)


jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -192,6 +192,21 @@
...
             --tenant-id ${UACLIENT_BEHAVE_AZ_TENANT_ID} || true
 
+- job:
+    name: uaclient-cleanup-stale-lxd-vms
+    defaults: ubuntu-advantage-client
+    node: torkoal
+    wrappers:
+      - workspace-cleanup
+    triggers:
+      - timed: 'H 1 * * *'
```
blackboxsw, line 202: This is good timing specification, can we also update the other lxd based tox behave-lxd and behave-vm jobs to be run at different hours than 1 instead of just "@daily" for those jobs too

lucasmoura: Thanks for that suggestion. I have updated the lxd jobs and the cleanup jobs as well

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -172,7 +172,7 @@
               credential-id: ua-azure-subscription-id
               variable: UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID
     triggers:
-      - timed: 'H 1 * * *'
+      - timed: 'H 0,4 * * *'
```
blackboxsw, line 175: are we trying to run this both at midnight an 4 am?, maybe we just want to clean it out at 0 to ensure we don't have stale stuff left over before starting our next run at 1:00 and 2:00

blackboxsw (changes requested):
suggested changes to make sure cleanup happens before lxd runs and azure

```diff
diff --git a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
index 00d441a..8721881 100644
--- a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
+++ b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
@@ -172,7 +172,7 @@
               credential-id: ua-azure-subscription-id
               variable: UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID
     triggers:
-      - timed: 'H 0,4 * * *'
+      - timed: 'H 0 * * *'
     builders:
       - shell: |
           set -ex
@@ -264,7 +264,7 @@
               credential-id: ua-contract-token-staging
               variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
     triggers:
-      - timed: 'H 2 * * *'
+      - timed: 'H 3 * * *'
     builders:
       - shell: |
           #!/bin/bash
@@ -292,7 +292,7 @@
               credential-id: ua-contract-token-staging
               variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
     triggers:
-      - timed: 'H 1 * * *'
+      - timed: 'H 2 * * *'
     builders:
       - shell: |
           #!/bin/bash
```

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
lucasmoura: My idea here is was to cleanup before and after the test are performed. We do this for AWS with two separate scrips, and I think it is worth doing for Azure as well. Otherwise we will have vms consuming resources for a day, which is unnecessary.

blackboxsw:
> suggested changes to make sure cleanup happens before lxd runs and azure
> 
> ```diff
> diff --git a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
> index 00d441a..8721881 100644
> --- a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
> +++ b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
> @@ -172,7 +172,7 @@
>                credential-id: ua-azure-subscription-id
>                variable: UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID
>      triggers:
> -      - timed: 'H 0,4 * * *'
> +      - timed: 'H 0 * * *'
>      builders:
>        - shell: |
>            set -ex
> @@ -264,7 +264,7 @@
>                credential-id: ua-contract-token-staging
>                variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
>      triggers:
> -      - timed: 'H 2 * * *'
> +      - timed: 'H 3 * * *'
>      builders:
>        - shell: |
>            #!/bin/bash
> @@ -292,7 +292,7 @@
>                credential-id: ua-contract-token-staging
>                variable: UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING
>      triggers:
> -      - timed: 'H 1 * * *'
> +      - timed: 'H 2 * * *'
>      builders:
>        - shell: |
>            #!/bin/bash
> ```

@lucasmoura cleared this up for me, azure cleanup job needs running twice because it cleans up all resources that match "uaclient" tag. so it should only wait until the current azure jobs have run.
as well as his 2 hour gap between lxd VM runs and lxd containers makes sense as VMs take a long time.

I retract me needs fixing comment.

PR #167, opened 2021-01-12
Add uaclient gcp test job to Jenkins

lucasmoura:
Add two jobs related to gcp in Jenkins. We are now introducing behave tests for generic gcp instances
and also a cleanup script to teardown instances that may not be deleted on the behave tests

### DIFF
--- a/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
+++ b/jenkins-jobs/ubuntu-advantage-client/jobs.yaml
@@ -205,7 +205,6 @@
 - job:
     name: uaclient-cleanup-stale-azure-vms
     defaults: ubuntu-advantage-client
-    node: torkoal
     wrappers:
       - workspace-cleanup
       - credentials-binding:
@@ -241,6 +240,32 @@
           python server-test-scripts/ubuntu-advantage-client/azure_cleanup.py --client-id ${UACLIENT_BEHAVE_AZ_CLIENT_ID} \
           --client-secret ${UACLIENT_BEHAVE_AZ_CLIENT_SECRET} --subscription-id ${UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID} \
             --tenant-id ${UACLIENT_BEHAVE_AZ_TENANT_ID} || true
+
+- job:
+    name: uaclient-cleanup-stale-gcp-vms
+    defaults: ubuntu-advantage-client
+    wrappers:
+      - workspace-cleanup
+      - credentials-binding:
+          - file:
+              credential-id: ua-gce-account
+              variable: GCP_CREDENTIALS_PATH
+    builders:
+      - shell: |
+          set -ex
+
+          git clone https://github.com/canonical/server-test-scripts.git --depth 1
+          git clone https://github.com/canonical/pycloudlib.git --depth 1
+          python3 -m venv gcp-cleanup
+          . gcp-cleanup/bin/activate
+          pip install pip --upgrade
+
+          cd pycloudlib
+          pip install -r requirements.txt
+          cd ..
+
+          https_proxy=http://squid.internal:3128 python server-test-scripts/ubuntu-advantage-client/gcp_cleanup.py \
+          --credentials-path $GCP_CREDENTIALS_PATH --project ubuntu-rickh --region us-west2 --zone a
 
 - job:
     name: uaclient-cleanup-stale-lxd-vms
@@ -447,6 +472,48 @@
           git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
           cd ua-client
           tox -e behave-azuregeneric-$UBUNTURELEASE
+
+- job-template:
+    name: uaclient-gcp-generic
+    defaults: ubuntu-advantage-client
+    project-type: matrix
+    execution-strategy:
+      sequential: true
+    axes:
+      - axis:
+          type: user-defined
+          name: UBUNTURELEASE
+          values: '{obj:ubunturelease}'
+    parameters:
+      - uaclient
+    wrappers:
+      - workspace-cleanup
+      - timeout:
+          timeout: 50
+          fail: true
+      - credentials-binding:
+          - text:
+              credential-id: ua-contract-token
+              variable: UACLIENT_BEHAVE_CONTRACT_TOKEN
+          - file:
+              credential-id: ua-gce-account
+              variable: UACLIENT_BEHAVE_GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H H * * *'
+    publishers:
+      - email-server-crew-qa
+      - trigger:
+          project: uaclient-cleanup-stale-gcp-vms
+          threshold: FAILURE
+    builders:
+      - shell: |
+          #!/bin/bash
+          set -ex
+
+          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
+          cd ua-client
+          no_proxy=launchpad.net https_proxy=http://squid.internal:3128 UACLIENT_BEHAVE_GCP_PROJECT=ubuntu-rickh \
+          tox -e behave-gcpgeneric-$UBUNTURELEASE
 
 - job-template:
     name: uaclient-aws-pro-{ubunturelease}
--- a/jenkins-jobs/ubuntu-advantage-client/default.yaml
+++ b/jenkins-jobs/ubuntu-advantage-client/default.yaml
@@ -41,6 +41,7 @@
       - uaclient-azure-pro
       - uaclient-aws-generic-{ubunturelease}
       - uaclient-azure-generic
+      - uaclient-gcp-generic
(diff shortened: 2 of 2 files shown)


jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +242,34 @@
...
+          git clone https://github.com/canonical/pycloudlib.git
+          python3 -m venv /tmp/gcp-cleanup
+          . /tmp/gcp-cleanup/bin/activate
+          pip install pip --upgrade
+
+          cd pycloudlib
+          pip install -r requirements.txt
+          cd ..
+
+          python server-test-scripts/ubuntu-advantage-client/gcp_cleanup.py --credentials-path $GCP_CREDENTIALS_PATH --project ubuntu-rickh --region us-west2 --zone a
```
paride, line 271: python -> python3

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +242,34 @@
...
+    defaults: ubuntu-advantage-client
+    node: torkoal
+    wrappers:
+      - workspace-cleanup
+      - credentials-binding:
+          - file:
+              credential-id: ua-gce-account
+              variable: GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H 0,4 * * *'
```
paride, line 256: Are we going to have other jobs starting GCP instances other than `uaclient-gcp-generic`? If that's the only job that will create instances to cleanup, then we can trigger this job when `uaclient-gcp-generic` ends  (with any state). The same as:

https://github.com/canonical/server-jenkins-jobs/blob/master/jenkins-jobs/cloud-init/integration-ec2.yaml#L27

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +242,34 @@
           --client-secret ${UACLIENT_BEHAVE_AZ_CLIENT_SECRET} --subscription-id ${UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID} \
             --tenant-id ${UACLIENT_BEHAVE_AZ_TENANT_ID} || true
 
+- job:
+    name: uaclient-cleanup-stale-gcp-vms
+    defaults: ubuntu-advantage-client
+    node: torkoal
```
paride, line 248: Let's not restrict this to torkoad (just drop this line). It should work on any node.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +242,34 @@
...
+              variable: GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H 0,4 * * *'
+    builders:
+      - shell: |
+          set -ex
+
+          git clone https://github.com/canonical/server-test-scripts.git
+          git clone https://github.com/canonical/pycloudlib.git
+          python3 -m venv /tmp/gcp-cleanup
```
paride, line 263: I think it's better if jobs keep their files in the Jenkins workspace when possible, so here I'd drop the `/tmp` and work in the PWD Jenkins puts us in (the job workspace).

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +242,34 @@
...
+              credential-id: ua-gce-account
+              variable: GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H 0,4 * * *'
+    builders:
+      - shell: |
+          set -ex
+
+          git clone https://github.com/canonical/server-test-scripts.git
+          git clone https://github.com/canonical/pycloudlib.git
```
paride, line 262: I think these are safe to `git clone --depth 1`.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -448,6 +476,39 @@
...
+      - workspace-cleanup
+      - timeout:
+          timeout: 50
+          fail: true
+      - credentials-binding:
+          - file:
+              credential-id: ua-gce-account
+              variable: UACLIENT_BEHAVE_GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H 3 * * *'
```
paride, line 502: If we go with the job trigger approach this timed trigger can be made `H H * * *` for more randomness.

paride (changes requested):

paride:
Let's wait for the squid change allowing `*.googleapis.com` to land before merging this. We'll likely need to set `https_proxy`.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
lucasmoura: Just to be sure, that job will be triggered even if the `ua-gce-generic` job fails, right ?

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
lucasmoura: I did not specify that here because the command is running inside a python3 virtualenv

lucasmoura:
@paride code updated

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
paride: Yes, that's the `threshold: failure` thing

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
paride: Right!

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -448,6 +472,43 @@
...
+      - timeout:
+          timeout: 50
+          fail: true
+      - credentials-binding:
+          - file:
+              credential-id: ua-gce-account
+              variable: UACLIENT_BEHAVE_GCP_CREDENTIALS_PATH
+    triggers:
+      - timed: 'H H * * *'
+    publishers:
```
paride, line 499: We also need to add an `email-server-crew-qa` publisher here, like https://github.com/canonical/server-jenkins-jobs/blob/master/jenkins-jobs/cloud-init/integration-ec2.yaml#L25.

We need this because this `publishers:` stanza replaces the publishers defined in `defaults.yaml` (where we already have `email-server-crew-qa`), and there's no way to tell JJB to append instead of replacing. I tried to find a workaround using yaml anchors, but it didn't really simplify things.

(The `email-server-crew-qa` sends emails to the broken mailing list, let's assume it will get fixed one day...)

paride (changes requested):
Added one comment, for the rest LGTM.

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
lucasmoura: Done

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -242,6 +241,31 @@
...
+          git clone https://github.com/canonical/pycloudlib.git --depth 1
+          python3 -m venv gcp-cleanup
+          . gcp-cleanup/bin/activate
+          pip install pip --upgrade
+
+          cd pycloudlib
+          pip install -r requirements.txt
+          cd ..
+
+          python server-test-scripts/ubuntu-advantage-client/gcp_cleanup.py --credentials-path $GCP_CREDENTIALS_PATH --project ubuntu-rickh --region us-west2 --zone a
```
paride, line 267: I added `https_proxy=http://squid.internal:3128` from the Jenkins interface and it worked :) Let's add it here, re-deploy the jobs to Jenkins from your branch, test them, and then I think this can land.

paride (changes requested):

jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -448,6 +473,44 @@
...
+          project: uaclient-cleanup-stale-gcp-vms
+          threshold: FAILURE
+    builders:
+      - shell: |
+          #!/bin/bash
+          set -ex
+
+          git clone --depth 1 -b "$UABRANCH" "$UAREPO" ua-client
+          cd ua-client
+          https_proxy=http://squid.internal:3128 UACLIENT_BEHAVE_GCP_PROJECT=ubuntu-rickh tox -e behave-gcpgeneric-$UBUNTURELEASE
```
paride, line 512: `launchpad.net` can't be accessed via proxy, I'm not sure why. Just add a `no_proxy=launchpad.net` here.

paride (approved):
All the jobs are green, let's merge this.

PR #173, opened 2021-02-19
[WIP] Add the pytest-based integration testing jobs

paride:


TheRealFalcon:
@paride , given how infrequently we push new stuff into ppas/repos/pockets, I don't see a lot of value in regularly testing things from these sources. I think the jobs that grab from repos should be manual, and we can kick them off ourselves when we push new things into the ppa or proposed. Another option would be a nightly job that checks if the version has been bumped, and then run tests only if it has

I think what we need from our nightly jobs is to simply git checkout/pull the latest code and run tox from master.

paride:
@TheRealFalcon thanks for looking at this! Let's leave the `-proposed` jobs aside for the moment, as they're only meant to run when there is a package in `-proposed`. The other jobs defined here test the daily builds of cloud-init (`ppa:cloud-init-dev/daily`) which are automatically built by Launchpad from git, so this is already a git checkout/pull, but this operation is done on Launchpad by the daily recipe and not by the jobs themselves.

We could of course re-build the daily package from git as part of these integration testing jobs, but I think we should use the daily builds we already have. Testing is what they're for after all.

Note that the *current* integration testing jobs (based on the cloud_tests) are basically doing the same:
1. grab the latest deb from the daily ppa (https://github.com/canonical/server-jenkins-jobs/blob/master/jenkins-jobs/cloud-init/integration-ec2.yaml#L107 via https://github.com/canonical/server-test-scripts/blob/master/cloud-init/daily_deb.sh#L27)
2. extract the git hash from the version string of the daily package (https://github.com/canonical/server-jenkins-jobs/blob/master/jenkins-jobs/cloud-init/integration-ec2.yaml#L114)
3. `git clone clout-init` and checkout the hash of the daily build (https://github.com/canonical/server-jenkins-jobs/blob/master/jenkins-jobs/cloud-init/integration-ec2.yaml#L118). This is necessary to make sure the test suite matches the package under test.

Here instead of cloning from git and checking out the right hash to match the daily package I made the the jobs use the .deb source package as a source for the tests too. The rationale is explained in this comment: https://github.com/canonical/server-jenkins-jobs/pull/173/files#diff-fed2e0bff0158c10e276021da200f6c4e03274c6f9052159febe8d0b0ce8fd46R69-R75. Comments are very welcome!

TheRealFalcon:
@paride :facepalm: Thanks, I forgot about the daily-ppa. That makes sense.

I think this is good. My only other comment on the current implementation is that I think it'd be easier to have "lxd_container" and "lxd_vm" for the platforms instead of "lxd" and "kvm". When looking at the jobs page it took me a minute to realize that the kvm jobs were the lxd vm jobs.

One other "future wishlist" idea I have is to be able to run a set of integration tests ad hoc against a branch for any particular platform or set of platforms. Certainly not necessary right now, but just mentioning it now in case that would impact how this is built.

paride:
The first suggestion is very easily done.

The second one requires some more work as we need to rebuild the .deb using the branch to test, but it's totally worth it. I'm at it.

PR #176, opened 2021-03-02
make uaclient-cleanup-stale-aws-vms more robust

paride:


jenkins-jobs/ubuntu-advantage-client/jobs.yaml
```diff
@@ -177,30 +177,25 @@
 - job:
     name: uaclient-cleanup-stale-aws-vms
     defaults: ubuntu-advantage-client
-    node: torkoal
     wrappers:
       - workspace-cleanup
     triggers:
-      - timed: 'H 4 * * *'
+      - timed: '@daily'
```
lucasmoura, line 183: I think we need the timed option here. The rationale behind is that we don't want to run the cleanup script while we have other aws tests running

paride: Right! I don't link this manual time synchronization of jobs, but to do this in a better way we need the postbuildscript jenkins plugin for which we have a pending RT...

lucasmoura (approved):

PR #179, opened 2021-03-18
Add the view definitions

paride:


paride:
Tested already. The `tableflip` view is renamed to `cloud-init` for consistency with all the rest. After all the uaclient jobs were already not under the `tableflip` view.

blackboxsw (approved):
LGTM tested cloud-init's regex and looks to match jobs fine

PR #182, opened 2021-04-02
Move the simplestreams jobs to their own view

paride:


paride:
No functional changes, this is only for clarity and uniformity. Test-deployed already.

mitechie (approved):
ty

PR #183, opened 2021-04-09
git-ubuntu: update nightly job to match MP CI

basak:
Commits 9fe8ea8, e79a3bc and fcfea1d are working well, so update the
nightly CI job the same way. This fixes the nightly CI job to work
correctly against the snapcraft core20 base that is now in master.


paride (approved):
LGTM, thanks!

PR #185, opened 2021-04-14
git-ubuntu: add ubuntu.com to no_proxy when doing snapcraft login

paride:


paride:
For some reason `snapcraft login` wants to contact `login.ubuntu.com` when logging in with the git-ubuntu credentials, while it does *not* do so with my personal credentials file, which I used for testing. I verified this in a clean VM environment and I can't tell for sure why the difference.

Let's just add `ubuntu.com` to `no_proxy` and be happy.

PR #188, opened 2021-04-29
ha virsh: go multi-release and add trigger

paride:


lucaskanashiro (approved):
Thanks @paride ! LGTM.

PR #190, opened 2021-05-14
cloud-init: add impish integration tests

orndorffgrant:
I went ahead and did a first pass at this, just following the patterns of past releases. I'm not sure how to test it though. Let me know if I missed anything.

### DIFF
--- a/jenkins-jobs/cloud-init/integration-ec2.yaml
+++ b/jenkins-jobs/cloud-init/integration-ec2.yaml
@@ -95,6 +95,22 @@
       - integration-ec2:
           release: hirsute
 
+- job:
+    name: cloud-init-integration-ec2-i
+    defaults: cloud-init
+    node: torkoal
+    triggers:
+      - timed: "H 0 * * 3"
+    publishers:
+      - email-server-crew-qa
+      - cloud-init-archive-results
+      - trigger:
+          project: cloud-init-integration-ec2-terminate
+          threshold: FAILURE
+    builders:
+      - integration-ec2:
+          release: impish
+
 - builder:
     name: integration-ec2
     builders:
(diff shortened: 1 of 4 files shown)


jenkins-jobs/cloud-init/integration-ec2.yaml
```diff
@@ -95,6 +95,22 @@
       - integration-ec2:
           release: hirsute
 
+- job:
+    name: cloud-init-integration-ec2-i
+    defaults: cloud-init
+    node: torkoal
+    triggers:
+      - timed: "H 0 * * 6"
```
paride, line 103: We run the `ec2` jobs on different days, so we don't exceed the limits of our ec2 account and to be sure that `cloud-init-integration-ec2-terminate` does not terminate instances that are still being used by another project.

Let's make the trigger:
```
timed: "H 0 * * 3"
```
as day 3 is not used by other ec2 jobs.

(The better way to do this is with the "throttle concurrent builds" plugin, which is installed but we need IS to create a job category for these ec2 jobs. At the moment I'm doing it for curtin, if it works fine I'll also ask for a cloud-init-ec2 category.)

paride (changes requested):
Thanks! There is just one little thing to change, see the inline comment.

jenkins-jobs/cloud-init/integration-ec2.yaml
orndorffgrant: Ah I didn't notice that, and thanks for the explanation! I'll make this change quick

orndorffgrant: @paride Just updated it!

paride: Thanks! Will merge after CI passes.

paride (approved):
LGTM!

PR #192, opened 2021-05-28
oci: add jobs for the standalone tests

paride:


jenkins-jobs/oci/unit.yaml
```diff
@@ -31,6 +31,48 @@
...
+          export https_proxy=http://squid.internal:3128
+          export no_proxy=launchpad.net,amazonaws.com,ecr.aws
+
+          git clone --depth 1 -b "$STSBRANCH" "$STSREPO" sts
+          cd sts/oci-unit-tests
+
+          DOCKER_PACKAGE={testcase} sh standalone-checks_test.sh
+    publishers:
+      - email:
+          recipients: server-crew-qa@lists.canonical.com sergio.durigan@canonical.com lucas.kanashiro@canonical.com
```
sergiodj, line 75: Can you remove @lucaskanashiro and add @athos-ribeiro and @bryceharrington here, please?

jenkins-jobs/oci/unit.yaml
```diff
@@ -31,6 +31,48 @@
...
+- job-template:
+    name: oci-unit-{testcase}-standalone
+    defaults: oci
+    triggers:
+      - timed: '@daily'
+    parameters:
+      - git-params-server-test-scripts
+    wrappers:
+      - timeout:
+          timeout: 5
```
sergiodj, line 45: This is "5 minutes", right?  I'm asking because sometimes these tests can take a while due to network latency and such.

sergiodj:
Thanks for the PR, @paride.  I'm leaving a couple of questions on it but it LGTM.  You can merge it after you address my request to update the email addresses :-).  Cheers!

jenkins-jobs/oci/unit.yaml
paride: Yes, from my testing 5 minutes seemed like enough, but we can bump this to something like 30 minutes. After all what we want to avoid here are "stuck forever" builds.

jenkins-jobs/oci/unit.yaml
paride: Done. I'd be +1 for adding the `ubuntu-docker-images` ML and dropping the individual names, but maybe let's get all the jobs in a better shape first, so Jenkins won't spam the list with false positives.

jenkins-jobs/oci/unit.yaml
sergiodj: +1, thanks.

PR #195, opened 2021-06-11
oci: update namespaces and provide aws region

paride:
- oci: skip the lts tests for non-lts packages
- oci unit: set AWS_PROFILE to allow logging in to public.ecr.aws


athos-ribeiro (approved):
Thanks, Paride :)

LGTM

PR #197, opened 2021-06-17
oci unit: add 'choice' parameter to chose the registry

paride:
When manually triggering the job this change will allow to chose the registry to pull the image from.


sergiodj:
LGTM, thanks @paride.

PR #199, opened 2021-06-25
ha: split the ha-agents test job in separate per-test jobs

paride:


paride:
@lucaskanashiro @basak I tested this change in Jenkins and it works nicely, however there is one caveat: `vm_utils.sh` in server-test-scripts calls the VM nodes like this:
```
VM01="fence-test-virsh-${UBUNTU_SERIES}-node01"
VM02="fence-test-virsh-${UBUNTU_SERIES}-node02"
VM03="fence-test-virsh-${UBUNTU_SERIES}-node03"
```
The test name is not part of the VM name, and should two split jobs run in parallel one jobs can interfere with the nodes of the other jobs. I think we should embed the agent under test in the VM name, something like:
```
VM01="ha-agent-virsh-${UBUNTU_SERIES}-${TEST}-node01"
VM02="ha-agent-virsh-${UBUNTU_SERIES}-${TEST}-node02"
VM03="ha-agent-virsh-${UBUNTU_SERIES}-${TEST}-node03"
```
This change should land in server-test-scripts.

There is another option: use Matrix jobs instead of separate jobs, and configure the matrix to run jobs serially and not in parallel. This is not difficult to do, but will make the test results less immediate, and in general matrix jobs are meant to test different configurations of the same thing, while here we're testing different things. Also retriggering a single test would become difficult. I'd keep the jobs separate and go for unique VM names.

lucaskanashiro:
@paride thanks for raising this, I think we should implement your first suggestion in the `server-tests-scripts`. I can work on that.

jenkins-jobs/ha/agents.yaml
```diff
@@ -15,17 +15,25 @@
...
+
+- project:
+    name: ha-agents-virsh
+    <<: *ha-agents-test-common
+    cloud:
       - virsh
+    test:
+      - fence_virsh
+      - resource_ipaddr2
+      - resource_systemd
```
lucaskanashiro, line 31: We are missing `resource_lvm-activate` here, it is already merged.

lucaskanashiro (changes requested):

jenkins-jobs/ha/agents.yaml
paride: Right!

paride:
@lucaskanashiro https://github.com/canonical/server-test-scripts/pull/118 landed and I added the missing agent, I think we can merge this, wdyt?

lucaskanashiro:
LGTM, +1. Feel free to merge it.

lucaskanashiro (approved):

PR #202, opened 2021-07-09
Handle tag conflicts in git-ubuntu MPs

basak:
If an MP submitter has a tag in their repo that collides with a tag in our main repo, then `git fetch --tags` against our main repo fails. We need the main repo tags to help `git describe` generate a suitable version string for the built snap. We don't need the submitter's tags at all, so we can just avoid fetching their tags and this avoids the issue.


paride (approved):
LGTM

PR #208, opened 2021-08-09
Add pre-commit configuration

paride:
In `tox.ini` replace the yamllint environment with a `pre-commit run --all-files` call, which will run several hooks including yamllint. This way running tox will still result in a full test run.

You can enable the pre-commit checks doing:
```
$ apt install pre-commit
$ pre-commit install # from withing this repository tree
```

This is in part a test to see how pre-commit works for us. Should we like it we can add a similar config to our other projects.


tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
-envlist = yamllint, jjb-test, jjb-lint
+envlist = pre-commit, jjb-test, jjb-lint
 skipsdist = true
 
-[testenv:yamllint]
-deps = yamllint==1.25.0
-commands = yamllint --strict .
+[testenv:pre-commit]
```
lucasmoura, line 5: This is just a nit, and you can ignore it if you think it is not a good idea, but maybe we could remove the `-` part here and just name it `precommit`

lucasmoura (approved):

PR #215, opened 2021-09-10
Revert "Fix cloud-init integration proposed"

paride:
The variable in curly brances is a JJB template variable, and it's never
seen by the shell.

This reverts commit e0f3e3d325e9ffb3a2a780938ae4063983ad9b27.


paride:
Hi @TheRealFalcon, this PR adds back the `$`, as I really think we don't want it there, however I'm sure there's an actual problem you were trying to fix! Was the job not finding the version in `-proposed`?

TheRealFalcon:
Hmmm, the lxd container jobs have rolled over at this point, but yes, I remember it was trying to execute
`rmadison --suite {release}-proposed cloud-init`
(without a substitution) and returning nothing. Adding the `$` made it work as I was expecting.

TheRealFalcon:
Hmmm, since we don't have the logs, we can't know for sure and my memory could be remembering something wrong. Sorry about that!

TheRealFalcon (approved):

paride:
No prob, let's see if the error comes back and how to deal with it!

PR #216, opened 2021-09-21
curtin: do not archive the cloud-init named pipe

paride:
Workaround for https://issues.jenkins.io/browse/JENKINS-28253


paride:
Deploying this should also redefine the `curtin-cloudinit-sru` job, which I modified to investigate the issue.

TheRealFalcon (approved):
Good find, thanks!

PR #217, opened 2021-09-24
c-i: built cloud-init-copr-build-stable-19.4 weekly

paride:
The branch is frozen, so daily builds aren't really necessary.


blackboxsw (approved):
Yes, reduce ongoing load for something that we don't really need to continuously get a signal for.

PR #219, opened 2021-10-05
c-i integration: add the credentials for GCP

paride:


paride:
I test-deployed this on Jenkins, but we need new cloud-init dailies to actually check if the jobs pass (because of the updated pycloudlib dep).

paride:
PR updated, most notably to set `https_proxy` and `no_proxy`.

PR #223, opened 2021-11-09
cii: allow for skipping non-lts releases on some platform

paride:
Releases can have a 'release-lts' flag that makes it easy to exclude
them using an 'exclude' stanza [1].

Also skip Jammy on Oracle for the moment as images are not available yet.
A better way would be checking for the availability of an image on a
given platform and mark the job as "unstable" (yellow) if there's no
image. This is however out of the scope of this change.

[1] https://docs.openstack.org/infra/jenkins-job-builder/definition.html#project


TheRealFalcon (approved):

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -17,19 +17,28 @@
 
 - project:
     name: cloud-init-integration-pytest
+    release:
+      - bionic:
+          release-lts: true
```
blackboxsw, line 22: bikeshed alert? would is-lts make more sense?
```suggestion
          is-lts: true
```
non-blocking comment

blackboxsw (approved):
This is really cool. thanks for pulling back the curtain and digging into the details on this selective job creation!

jenkins-jobs/cloud-init/integration.yaml
paride: I agree; I'll go for `release-is-lts` as I find it more explicit. Thanks for reviewing!

PR #224, opened 2021-11-11
Update triage email for separate cloud-init and UA triagers (SC-597)

TheRealFalcon:
Not a whole lot to test with, so this could use a thorough review.


jenkins-jobs/daily-triage/tableflip-triage-email.sh
```diff
@@ -23,27 +23,34 @@ github_projects=(cloud-init pycloudlib)
...
-    ndays=3
-    triager=${triagers[$((week%4 - 1))]}
+    cloud_init_triager=${cloud_init_triagers[$((triager_index % cloud_init_triager_count))]}
+    ua_triager=${ua_triagers[$((triager_index % ua_triager_count))]}
 else
-    ndays=1
-    triager="nobody"
+    ndays=3
+    cloud_init_triager="nobody"
+    ua_triager="nobody"
```
paride, line 51: I am not sure I get why "nobody" on Mondays. We still want somebody to do triage on Mondays for the weekend + previous Friday, right?

jenkins-jobs/daily-triage/tableflip-triage-email.sh
```diff
@@ -23,27 +23,34 @@ github_projects=(cloud-init pycloudlib)
...
 week=$(date --utc '+%-V')
 dow=$(date --utc '+%u')
-if [[ "$dow" -ge 2 && "$dow" -le 5 ]]; then
+
+# Bash is the worst...
+# The point here is to determine the current day of the year
+# excluding weekends. E.g., if day 1 is a Monday, day 5 is Friday,
+# the following Monday is day 6. We can then use this number to
+# mod by the number of triages to cycle our triagers throughout the year.
+# Sorry to anybody who might get double triage on January 1st ;)
```
paride, line 40: I think the logic here is "fair" but has a downside: the triage weekday rotates, while in the past we tried to keep the triage day on a fixed weekday (a part for Mondays, where 3 days are triaged, so it's unfair to have the same person always do triage on Monday). Having a fixed triage day makes it easier to remember, but I can see it's not so easy for how the squad is organized at the moment. Let's bring this up at standup and see if we all agree with the change.

jenkins-jobs/daily-triage/tableflip-triage-email.sh
```diff
@@ -112,21 +120,6 @@ subject="Daily triage for: ${projects[*]} [$triager]"
...
-        echo "$d: ${triagers[$i]}"
-        i=$((i+1))
-    done
-    printf '\nMondays follow the same schedule, starting from\nthe first Monday of the year. Next Mondays:\n\n'
-    for i in {1..5}; do
-        future_date=$(date --utc --date="$i Monday" '+%b %_d')
-        future_week=$(date --utc --date="$i Monday" '+%-V')
-        future_triager=${triagers[$((future_week%4 - 1))]}
-        echo "$future_date: $future_triager"
-    done
```
paride, line 129: This was mostly useful to make the Monday triage rotation less a surprise, given that it "rotates". I'm fine with dropping this is we decide we're file with always rotating.

paride (commented):
Thanks James! I left some inline comments, I think we need we need another short discussion at standup before landing this change.

jenkins-jobs/daily-triage/tableflip-triage-email.sh
paride: Right, I tested it using `faketime 2021-11-01` (a Monday) but that's still counts as Sunday because the time it `00:00:00`.  With `faketime 2021-11-01T06:00` it works as expected, but we're not triaging 3 days anymore on Mondays, while we need to triage the Friday + the weekend...

jenkins-jobs/daily-triage/tableflip-triage-email.sh
```diff
@@ -23,27 +23,34 @@ github_projects=(cloud-init pycloudlib)
...
+# The point here is to determine the current day of the year
+# excluding weekends. E.g., if day 1 is a Monday, day 5 is Friday,
+# the following Monday is day 6. We can then use this number to
+# mod by the number of triages to cycle our triagers throughout the year.
+# Sorry to anybody who might get double triage on January 1st ;)
+triager_index=$((week * 5 + dow))
+cloud_init_triager_count=${#cloud_init_triagers[@]}
+ua_triager_count=${#ua_triagers[@]}
+if [[ "$dow" -ge 1 && "$dow" -le 5 ]]; then
     ndays=1
```
blackboxsw, line 45: You've change the logic a bit now. On Mondays we want ndays to be 3 to account for the two weekend days.

I think we want the following diff so the Monday triagers get 3 days worth of bug visibility and every other day only reports ndays=1
```diff
diff --git a/jenkins-jobs/daily-triage/tableflip-triage-email.sh b/jenkins-jobs/daily-triage/tableflip-triage-email.sh
index 0ccd8c1..6faa1fb 100755
--- a/jenkins-jobs/daily-triage/tableflip-triage-email.sh
+++ b/jenkins-jobs/daily-triage/tableflip-triage-email.sh
@@ -38,17 +38,18 @@ dow=$(date --utc '+%u')
 # the following Monday is day 6. We can then use this number to
 # mod by the number of triages to cycle our triagers throughout the year.
 # Sorry to anybody who might get double triage on January 1st ;)
-triager_index=$((week * 5 + dow))
+triager_index=$((week * 5 + dow ))
 cloud_init_triager_count=${#cloud_init_triagers[@]}
 ua_triager_count=${#ua_triagers[@]}
+ndays=1
+cloud_init_triager="nobody"
+ua_triager="nobody"
 if [[ "$dow" -ge 1 && "$dow" -le 5 ]]; then
-    ndays=1
+    if [ $dow -eq 1 ]; then
+        ndays=3
+    fi
     cloud_init_triager=${cloud_init_triagers[$((triager_index % cloud_init_triager_count))]}
     ua_triager=${ua_triagers[$((triager_index % ua_triager_count))]}
-else
-    ndays=3
-    cloud_init_triager="nobody"
-    ua_triager="nobody"
 fi
 
 # Retrieve the bugs
```

blackboxsw (changes requested):

jenkins-jobs/daily-triage/tableflip-triage-email.sh
paride: However as Chad noted we still need to treat Mondays in a special way, as on Mondays we have to triage the previous 3 days (weekend + Friday).

TheRealFalcon: Thanks, I think I have that now.

paride (approved):
I tested this a bit with commands like `TRIAGE_NO_SENDMAIL=1 faketime 2021-11-01T06:00:00 ./tableflip-triage-email.sh` and it LGTM, seems to behave well on Mondays, non-Monday-weekdays and on weekends (when we're not running it as it's not scheduled in Jenkins). There's a small annoyance, this warning when querying for UA project bugs:
```
warn: no Launchpad project with that name exists
```
as there's no `ubuntu-advantage-client` Launchpad project. This was born as a simple script to query LP for bugs, but now want a summary of different info coming for heterogeneous sources, and the approach is showing its limits. Let's stand the warning for now and see if we can do better in Jira.

I'm approving but not merging for now, in case Chad wants to have another look. Let's remember to squash the commit history when landing this PR.

blackboxsw (approved):
LGTM!

PR #226, opened 2021-12-06
Use yaml anchors to make jobs inherit the global defaults

paride:
This reduces duplication in the 'defaults' stanzas. For this change we
need to set:

  retain_anchors=True

in jenkins_jobs.ini, which in general will allow us to make more use of
anchors and aliases.


blackboxsw (approved):
YAML anchors FTW.  In reading up on [retain_anchors JJB config feature](https://docs.openstack.org/infra/jenkins-job-builder/execution.html?highlight=retain_anchors#job-build) It says
```
(Optional) If set to True, YAML anchors will be retained across files, allowing jobs to be 
composed from bits of YAML defined in separate files. Note this means that the order of 
processing files matters - jenkins-jobs loads files in alphabetical order (all files in a dir are 
loaded before any files in subdirs). For example, if your anchors are in a file named foo.yml 
they will be accessible in qux.yml but not in bar.yml. They will also be accessible in 
mydir/bar.yml and mydir/qux.yml. False by default) 
```


So, thankfully our defaults are stored in _common/ which is ordered lexigraphically before any of our other job definitions. Looks good to me, just something we need to be aware of if we try to source common anchors/aliases in a subdirectory that is ordered before _common :)

PR #229, opened 2022-01-20
cii: refactor, add jobs doing IN_PLACE testing from git (SC-745)

paride:
Also:
 - Refactor jobs definitions
 - Drop Hirsute


paride:
I want to test a few job runs before declaring this as ready to be merged, but we should be quite close.

paride:
Note that we can't run the integrations tests on all the platforms from git, as `CLOUD_INIT_SOURCE=IN_PLACE` only works with LXD.

paride:
Also note that the jobs are not limited to `torkoal` anymore, which is good as it means they're not bound anymore to the config of a specific node. The pycloudlib conf file made this easier.

paride:
I think the whole LXD thing to get the daily/proposed deb can be replaced by `pull-pkg`:
```
pull-lp-source cloud-init focal-proposed
```
```
pull-ppa-source --ppa ppa:cloud-init-dev/daily cloud-init focal
```

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -35,24 +34,67 @@
...
+
+- builder:
+    name: cloud-init-integration-get-source-from-daily-ppa
+    builders:
+      - shell: |
+          #!/bin/bash
+
+          set -eux -o pipefail
+
+          pull-ppa-source --ppa=cloud-init-dev/daily --no-verify-signature cloud-init {release}
```
blackboxsw, line 70: TIL about pull-ppa-source. thanks for this gem @paride . Do we have config/anisble logic to guarantee that ubuntu-dev-tools is installed on the jenkins nodes?

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -23,7 +23,6 @@
           release-is-lts: true
       - focal:
           release-is-lts: true
-      - hirsute
```
blackboxsw, line 26: good cleanup.

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -35,24 +34,67 @@
...
+          pull-lp-source cloud-init {release}
+          ln -s cloud-init-*/ cloud-init
 
 - job-template:
     name: cloud-init-integration-{platform}-{release}-{source}
     defaults: cloud-init
-    node: torkoal
     triggers:
-      - timed: !j2: "{{ '@daily' if platform == 'lxd_container' else '@weekly' }}"
+      - timed: !j2: "{{ {'git': '@daily', 'daily': '@weekly'}.get(source, '') }}"
```
blackboxsw, line 92: @paride please confirm what this line means here. 
To me it looks like we extract "@daily" if source == 'git', '@weekly' if source == "daily" or '' if neither. A value of '' means the job will not have a timer, but can be manually kicked right?

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -35,24 +34,67 @@
       - gce
       - oci:
           platform-lts-only: true
+    source:
+      - git
+      - daily
```
blackboxsw, line 39: bikeshed, because of my confusion on timer jobs "daily" == "@weekly" is it worth just making this value "daily-ppa" instead to be more clear below.

blackboxsw (commented):
Looks slick @paride. I have a couple of questions but this is a great cleanup.

jenkins-jobs/cloud-init/integration.yaml
paride: Those `pull-pkg` tools are super useful, but sometimes it seems that they're hidden in plain sight, and we forget they exist. Anyway yes, `ubuntu-dev-tools` is in the Ansible config.

jenkins-jobs/cloud-init/integration.yaml
paride: Yes exactly. I verified that Jenkins accepts empty `timed` trigger and behaved well, it does.

jenkins-jobs/cloud-init/integration.yaml
paride: Agreed, that's confusing, `-daily-ppa` is much better.

paride: More bikeshedding: I'll also rename the `-proposed` jobs to `-archive`. We'll just have to remember that they always run from the tip of the archive, including `-proposed`, if there's a newer package available there.

When doing SRU verification it's important to check that the tested package version is the one we're interested in, but this is something that had to be done anyway.

This change will allow for some more simplification of the definitions.

paride:
I'm unhappy with the fact that the git parameters are defined in all the jobs, even the non-git ones (`-daily-ppa`, `-proposed`), where they are meaningless. I tried adding Jinja2 templating so that the parameters are only set if `{source} = git`, but apparently JJB only supports rendering j2 templates *as strings*, not as yaml.

Interestingly JJB merged support for rendering j2 templates as yaml at come point, but the change was reverted in 2019 by OddBloke: https://review.opendev.org/c/jjb/jenkins-job-builder/+/673347 (it's a small world!).

paride:
@blackboxsw I manually deployed this to Jenkins for easier testing, so you can have a look at the actual jobs if you want.

blackboxsw (approved):
LGTM. This is much cleaner and thanks for syncing the jobs over into jenkins for testing it aided my checks on template -> jobname rendering.

PR #230, opened 2022-02-04
c-i-i: drop the -git jobs

paride:
Testing the git source directly without building a .deb package requires
IN_PLACE testing, which is mostly meant to be a development tool and not
really useful to have in automation.

Testing jobs from git can still be done by triggering a daily build on
Launchpad and then triggering one of the -daily jobs.


paride:
Thanks @TheRealFalcon and @blackboxsw for the discussion on this one.

TheRealFalcon (approved):
Thanks!

PR #231, opened 2022-02-04
Switch CentOS 8 to CentOS Stream 8

holmanb:
```
Centos 8 went EOL over a month ago, and it looks like this job is
failing due to repo issues. The job passes locally when using CentOS
Stream 8, switch to that.

Also add CentOS Stream 9. This test currently fails (ignore for now).
The fix requires updating tools in cloud-init which select dependencies
(package epel-release changes), see here[1] for more details:

[1] https://docs.fedoraproject.org/en-US/epel/
```

If you would rather we fix the cloud-init tooling prior to merging the CentOS Stream 9 test, I can do that, just let me know.


TheRealFalcon (commented):
LGTM, but let's wait and let Paride comment.

paride:
Thanks @holmanb, unfortunately this isn't enough to get the copr builds/tests passing again. We also need to bump the base centos version we use to build the source package `cloud-init-copr-build`, but even that is not quite enough: the build fails on the following chroots:

- `centos-stream-8-x86_64`
- `epel-9-x86_64` (available already)

due to missing dependencies. I'm digging a bit deeper, I think we'll need to make other adjustments here and in some build scripts.

holmanb:
> Thanks @holmanb, unfortunately this isn't enough to get the copr builds/tests passing again. We also need to bump the base centos version we use to build the source package `cloud-init-copr-build`, but even that is not quite enough: the build fails on the following chroots:
> 
>     * `centos-stream-8-x86_64`
> 
>     * `epel-9-x86_64` (available already)
> 
> 
> due to missing dependencies

I'm not sure where the change needs to be made, but here's a quick overview of what I've found:

Based on the [copr build failure logs](https://copr.fedorainfracloud.org/coprs/g/cloud-init/cloud-init-dev/build/3390934/), it looks like the missing deps are `python3-httpretty` (on both stream-8 and epel-9), and `python3-pytest-cov` (on epel-9). I fired up a centos stream 8 container and it looks like installing the package `epel-release` enables the repos required to install these deps. That said, I don't know where the copr chroot scripts/configurations are managed, so I can't propose any changes at this time. Hopefully that helps.

paride:
OK, here is my take on this. Similarly to building a `.deb` building a `.rpm` is a two stage process:

1. A source pacakge is built from upstream source + packaging data
2. The source file it built to produce a binary rpm.

Step 1 is done in a LXD container by the `cloud-init-copr-build` job, not touched by this PR but same file. Currently we set `centos_version: 8` there, which should become `centos_version: 8-Stream`, as `centos/8` is not available anymore in the `images:` LXD remote. Note that we can build the source package 8-Stream and then build the binary packages for a newer release.

Step 2 is done by uploading the `.src.rpm` file to COPR, which has a number of build chroots enabled to build binary packages on/for different releases. Currently we have:

- `epel-8-x86_64`
- `fedora-rawhide-aarch64`
- `fedora-rawhide-ppc64le`
- `fedora-rawhide-x86_64`

I *think* `epel-8-x86_64` currently means CentOS 8-Stream + EPEL 8, which a target we want to keep. Available is also:

- `epel-9-x86_64` (Pre-release builds, against CentOS Stream 9 + EPEL 9, this will be changed once RHEL 9 is available.)

I'm not sure of what they mean by "this will be changed", but I tried enabling it and the builds fails because of missing dependencies not available in EPEL 9 (at least: `python3-httpretty`). The deps are really missing for the moment (I tried to guess python3/python renames to no avail), and I think there's little we can do about this, which means we're not going to build on 9-Streams for now.

If we agree that we can't build on CentOS 9 + EPEL for now then we can put this PR in standby (or close it for now) and fix the current build with these two changes:

- https://github.com/canonical/server-jenkins-jobs/pull/233
- https://github.com/canonical/server-test-scripts/pull/148

and live with 8-Stream + EPEL for now.

holmanb:
@paride: I'm fine with not including CentOS Stream 9 until it is capable of running tests. I have commented on your PRs and I can close this one in favor of yours.

PR #234, opened 2022-02-11
admin-jobs-update: allow skipping reading/writing the JJB cache entirely

paride:


paride:
@blackboxsw this should allow to work on in-development jobs without disabling `admin-jobs-update` without resorting to disabling the job manually (prone to forgetting it disabled, as it happened to me already!).

paride:
Useful reference about JJB usage of `XDG_CACHE_HOME`: https://docs.openstack.org/infra/jenkins-job-builder/execution.html#f1.

PR #235, opened 2022-02-16
c-i-i: run the daily-ppa jobs weekly/2 and add trigger for *-proposed

paride:


paride:
The`cloud-init-integration-manual-proposed-trigger` job will of course trigger a lot of jobs. This shouldn't be a problem jobs running on real clouds, but the LXD based jobs can put some heavy load on the Jenkins agent nodes. However every node has a limited number of jobs it can executed, if we end up overloading the node we should maybe lower that.

The idea is of course to trigger all the -proposed jobs at SRU time with one click.

paride:
Reviewed in person (Frankfurt sprint).

PR #236, opened 2022-02-22
HA: only test agents on supported releases

paride:


paride:
@lucaskanashiro let me know if I forgot to exclude any unsupported combination.

PR #238, opened 2022-03-07
c-i-i: major refactoring and conversion to pipeline jobs (SC-875)

paride:
User visible changes:
 * Drop the -proposed jobs; make the .deb source a parameter
 * Add -trigger jobs, watching the daily-ppa for changes
 * Allow to limit which tests to run
 * Add -ALL manual trigger jobs
 * Move the list of Ubuntu releases to _common
   (to be used by other projects in the future).


paride:
After merging this we'll want to run `admin-jobs-update` with `delete_old=true` to get rid of the other (old and test) jobs.

jenkins-jobs/cloud-init/integration.yaml
```diff
@@ -17,112 +17,285 @@
...
+
+            steps {{
+              dir('cloud-init') {{
+                sh 'tox -e integration-tests-jenkins -- -v --junitxml="$WORKSPACE/junit-report.xml" --showlocals "$pytest_file_or_dir"'
+              }}
+            }}
+
+            post {{
+              cleanup {{
+                sh 'echo TODO'
```
TheRealFalcon, line 174: Were you expecting anything to be here before we merge this?

paride: Thanks for reviewing! The TODO is not for this PR: my plan it to add an extra cleanup stage to be really sure that no stale instances are left running in the clouds. Currently we don't have this, and this PR doesn't implement it either, but that's the place for it.

PR #239, opened 2022-03-09
Temporarely disable the email publishers

paride:
Waiting for RT: #136241 to be addressed.


paride:
Apparently the failure to send emails doesn't cause the jobs to be marked as FAILURE, we "only" get an ugly traceback in the console log. So maybe maybe we can avoid merging this, however I'll keep it up in case the broken email setup is causing other issues which I don't immediately see.

athos-ribeiro (approved):
LGTM as is. +1 to holding this for now since it is not causing failures.

